PR #66, opened 2021-07-26
`BackgroundColor` to `Background` for Customizing Controls

jfversluis:
I think the `IFrameworkElement.BackgroundColor` got renamed to `IFrameworkElement.Background`, updated that.

Additionally; for the removal of the underline of an entry example. Is it supposed to say Background? @davidbritch 
It seems kind of funny, doesn't it then only work when a background is set?

### DIFF
--- a/docs/user-interface/handlers/customize.md
+++ b/docs/user-interface/handlers/customize.md
@@ -38,7 +38,7 @@ public partial class App : Application
         InitializeComponent();
 
 #if __ANDROID__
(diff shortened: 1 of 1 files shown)


davidbritch:
Hi @jfversluis 

Thanks for the BackgroundColor vs Background fix.

Ultimately you can say specify any string as the index into the mapper. So you could have `Handlers.EntryHandler.EntryMapper["Gerald"]` and then update any native properties you want.

Eventually the doc will be updated/replaced with full explanations.

PR #122, opened 2021-08-18
Customize handlers

davidbritch:
A much expanded doc on how to customize handlers.


docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+All handler-based .NET MAUI controls can be customized by modifying the mapper for a control's handler. The syntax for modifying a mapper is as follows:
+
+```csharp
+HandlerType.MapperName[string] = (handler, view) =>
+{
+  // Manipulate handler.NativeView to customize the handler
+};
+```
+
+All handler types are located in the `Microsoft.Maui.Handlers` namespace, and each handler provides a mapper that maps each cross-platform control property to a method that applies the property value to the native control. The `handler` argument to the mapper provides an instance of the handler being customized, and the `view` argument provides an instance of the cross-platform control that the handler implements. For a list of the type names that implement handler-based .NET MAUI controls, see [Handler-based controls](#handler-based-controls).
```
adegeo: I think there was a minor bug here. Perhaps I fixed it?
```suggestion
All handler types are located in the `Microsoft.Maui.Handlers` namespace, and each handler provides a mapper that maps each cross-platform control property to a method that applies the property value to the native control. The `handler` argument provides an instance of the handler being customized, and the `view` argument provides an instance of the cross-platform control that the handler implements. For a list of the type names that implement handler-based .NET MAUI controls, see [Handler-based controls](#handler-based-controls).
```

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
 
-Handlers are global and therefore customization can occur anywhere in your .NET MAUI app. For example, customizing a handler for a control in your `App` class will result in all controls of that type being customized in the app. Similarly, customizing a handler for a control in the first page of your app will also result in all controls of that type in the app being customized.
+Handlers can be customized to augment the appearance and behavior of a .NET MAUI control beyond the customization that's possible through the control's API. Handlers are global and therefore customization can occur anywhere in your .NET MAUI app. For example, customizing a handler for a control in your `App` class will result in all controls of that type being customized in the app. Similarly, customizing a handler for a control in the first page of your app will also result in all controls of that type in the app being customized.
```
adegeo: UPDATE on this comment after reading further:

This suggestion may not make sense. After reading further I think perhaps my understanding about how this all works was wrong and therefore my suggestion is invalid. I'll leave it here though.

----

Maybe be more direct instead of using examples to demonstrate the concept? Maybe you want to tweak this further or ignore it? Just edit my comment here. Is **view** the right term? I'm unsure...
```suggestion
Handlers can be customized to augment the appearance and behavior of a .NET MAUI control beyond the customization that's possible through the control's API. Handler customizations are global and aren't scoped to a specific UI view. Handler customization is allowed to happen anywhere in your app. Once a handler is customized, it affects all native controls of that type, everywhere in your app.
```

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
 
-Handlers are typically customized to augment the appearance and behavior of a native platform control. Handlers can be customized per platform by using the `#ifdef` compiler directive, to multi-target code based on the platform. However, you can just as easily use platform-specific folders or files to organize such code.
+## Handler lifecycle
 
-> [!NOTE]
-> Handler customization can range from the simple to the complex. More complex examples will appear in time.
+All handler-based .NET MAUI controls support two handler lifecycle events:
 
-### Customize the background color of all controls
+- `HandlerChanged` is fired after the handler for a cross-platform control has been created. This event indicates that the native control that implements the cross-platform control is available, and all the property values set on the cross-platform control have been applied to the native control.
```
adegeo: You've used `fired` and `fires` a lot 😀 but the proper term for events is `raised`. The code raises the event or the event was raised. Code doesn't fire events and the event wasn't fired.

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+- `HandlerChanging` is fired before an existing handler is removed from a cross-platform control, and before a new handler for the cross-platform control is created. This event indicates that the existing native control is about be removed from the cross-platform control, and therefore that any native events should be unwired and other cleanup performed. The `HandlerChangingEventArgs` object that accompanies this event has `OldHandler` and `NewHandler` properties, of type `IElementHandler`.
```
adegeo: I would be clear on the `HandlerChanging` and `HandlerChanged` events. to me it sounds like this is true and may be worth calling out (especially if I'm wrong in these assumptions):

01. The `HandlerChanging` event is raised before the `HandlerChanged` event.
01. This doesn't happen the first time handlers are added to controls, since there is nothing being "changed" (removed before an add).

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
 
-The following example customizes the background color of every control in the app, on Android, to cyan:
+In addition to these events, each control also has an overridable `OnHandlerChanged` method that's invoked when the `HandlerChanged` event fires, and a `OnHandlerChanging` method that's invoked when the `HandlerChanging` event fires.
+
+## Customize a control with a mapper
+
+The .NET MAUI `Entry` is a single-line text input control, that implements the `IEntry` interface. On iOS, the `EntryHandler` maps the `Entry` to an iOS `UITextField`. On Android, the `Entry` is mapped to an `AppCompatEditText`, and on Windows the `Entry` is mapped to a `TextBox`:
```
adegeo: I think it helps to say control here:
```suggestion
The .NET MAUI `Entry` is a single-line text input control, that implements the `IEntry` interface. On iOS, the `EntryHandler` maps the `Entry` to an iOS `UITextField` control. On Android, the `Entry` is mapped to an `AppCompatEditText`, and on Windows the `Entry` is mapped to a `TextBox` control:
```

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+                handler.NativeView.FontWeight = Microsoft.UI.Text.FontWeights.Thin;
 #endif
+            };
+        }
     }
 }
 ```
 
-### Remove Android Entry underline
+In this example, the `Entry` customization occurs in a page class. Therefore, all `Entry` controls on Android, iOS, and Windows will be customized once the `CustomizeEntryPage` is displayed. The following customization is performed by using compiler preprocessing directives:
```
adegeo: Am I correct in this assumption? On the other UI stacks I've used, you have to call Show or add the page to something else. Technically you could do `var page = new CustomizeEntryPage()` and that doesn't show the page, but the constructor has run and now the handler code has run.

```suggestion
In this example, the `Entry` customization occurs in a page class. Therefore, all `Entry` controls on Android, iOS, and Windows will be customized once an instance of the `CustomizeEntryPage` class is created, but before it's displayed. The following customization is performed by using compiler preprocessing directives:
```

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+
+- On Android, the underline is removed from the `Entry`.
```
adegeo, line 88: Is this correct? The code looks like it's making the background transparent? (unless the background is controlling the visibility of the underline?)

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+- On iOS, the border is removed from the `Entry`.
+- On Windows, the thickness of the font in the `Entry` is set to thin.
+
+> [!IMPORTANT]
+> The index into a mapper is an arbitrary string and does not have to correspond to a property exposed by a type, or a property on a native control. For example, `["MyCustomization"]` can be specified as a mapper index, with any native control modification being performed as the customization.
```
adegeo:
```suggestion
> The index into a mapper is an arbitrary string and doesn't have to correspond to a property exposed by a type, or a property on a native control. For example, `["MyCustomization"]` can be specified as a mapper index, with any native control modification being performed as the customization.
```

adegeo: Question on this. The "MyCustomization" doesn't make sense in this note and in the previous code example. How does that map to an `Entry` control?

**UPDATE**
Looking at the code below in the next section, there is a type check: `if (view is MyEntry)` so this makes me think that ALL handlers are applied to every control when they're created. There is no system that automatically runs handler X when control X is created. I didn't get that impression until I got this far and the IMPORTANT note confused me and then I saw the following code. That may prove confusing.

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+#elif __iOS__
+        handler.NativeView.BorderStyle = UIKit.UITextBorderStyle.None;
+#elif WINDOWS
+        handler.NativeView.FontWeight = Microsoft.UI.Text.FontWeights.Thin;
+#endif
+    }
+};
+```
+
+Any `MyEntry` instances in the app will then be customized as per the handler modification.
```
adegeo, line 128: One thing I'm not clear about, do handlers affect already created controls? Or just apply to any control created after the handler is altered?

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+                nativeView.SetBackgroundColor(Colors.LightPink.ToNative());
+            else
+                nativeView.SetBackgroundColor(Colors.White.ToNative());
         }
     }
 }
 ```
 
-Any `MyEntry` instances will then be customized as per the handler modification.
+The advantage of this approach is that compiler preprocessing directives aren't required, and that the partial methods don't have to be implemented on each platform. Instead, the compiler removes references to the method signature on any platforms that don't provide an implementation. For information about partial methods, see [Partial methods](/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods#partial-methods).
```
adegeo: > the compiler removes references to the method signature on any platforms that don't provide an implementation.

Is this something special to MAUI? I've never heard of this before. So when compiling for iOS the compiler would remove the `ChangedHandler(sender, e);` call instead of erroring out that a `ChangedHandler` method doesn't exist?

docs/user-interface/handlers/customize.md
```diff
@@ -1,122 +1,316 @@
...
+- `FlyoutPage`
+- `IndicatorView`
+- `ListView`
+- `Map`
+- `RefreshView`
+- `SwipeView`
+- `TabbedPage`
+- `TableView`
+
+For more information about renderers, see [Xamarin.Forms custom renderers](/xamarin/xamarin-forms/app-fundamentals/custom-renderer/).
```
adegeo: Xamarin link here?

adegeo (approved):
Lots of questions and suggestions on here 😁

docs/user-interface/handlers/customize.md
davidbritch: (2) is wrong so I'll clarify.

docs/user-interface/handlers/customize.md
davidbritch: There was an UnderlineVisible property in an earlier preview, which seems to have vanished, so I had to make the background transparent to remove the underline drawable. I'll take another look but this may be as good as it gets now.

davidbritch: Setting the background color to transparent is the way.

docs/user-interface/handlers/customize.md
davidbritch: I *think* this is resolved from an earlier comment that you went back and edited.

docs/user-interface/handlers/customize.md
davidbritch: Removed.

docs/user-interface/handlers/customize.md
davidbritch: I've re-written this, but yes, there's no error about `ChangedHandler` not existing if it's not implemented on a platform.

The .NET partial method doc says (about partial methods) "If the developer does not supply an implementation, the compiler removes the signature at compile time.".

davidbritch:
Replaced by #471

PR #144, opened 2021-08-27
[essentials] More essentials articles

adegeo:
There are a few TODOs to check out. The permissions article has some structural changes that I felt made more sense (info first, then examples, instead of examples, info, examples)

### DIFF
--- a/docs/platform-integration/essentials/includes/android-permissions.md
+++ b/docs/platform-integration/essentials/includes/android-permissions.md
@@ -3,6 +3,8 @@ ms.topic: include
 ms.date: 08/04/2021
 ---
 
+### Android permissions
+
 This API uses runtime permissions on Android. Please ensure that .NET MAUI Essentials is fully initialized and permission handling is setup in your app. For more information, see [Get Started with .NET MAUI Essentials](../get-started.md).
 
 In the Android project's `MainLauncher` or any `Activity` that is launched, .NET MAUI Essentials must be initialized in the `OnCreate` method:
--- a/docs/platform-integration/essentials/maps.md
+++ b/docs/platform-integration/essentials/maps.md
@@ -1,45 +1,42 @@
 ---
-title: "Xamarin.Essentials Map"
-description: "The Map class in Microsoft.Maui.Essentials enables an application to open the installed map application to a specific location or placemark."
-ms.date: 05/26/2020
+title: "Map"
+description: "Learn how to use the .NET MAUI Map class in the Microsoft.Maui.Essentials namespace. This class enables an application to open the installed map application to a specific location or placemark."
+ms.date: 08/24/2021
 no-loc: ["Microsoft.Maui", "Microsoft.Maui.Essentials"]
 ---
 
 # Map
 
-The `Map` class enables an application to open the installed map application to a specific location or placemark.
+This article describes how you can use the .NET Multi-platform App UI (.NET MAUI) Essentials `Map` class. This class enables an application to open the installed map application to a specific location or placemark.
 
 ## Get started
 
 [!INCLUDE [get-started](includes/get-started.md)]
 
-## Using Map
-
 [!INCLUDE [essentials-namespace](includes/essentials-namespace.md)]
 
-The Map functionality works by calling the `OpenAsync` method with the `Location` or `Placemark` to open with optional `MapLaunchOptions`.
+## Using the map
+
+The map functionality works by calling the `Map.OpenAsync` method, and passing either an instance of the `Location` or `Placemark` type. The following example opens the installed map app at a specific GPS location:
 
 ```csharp
-public class MapTest
+public async Task NavigateToBuilding25()
 {
-    public async Task NavigateToBuilding25()
-    {
-        var location = new Location(47.645160, -122.1306032);
-        var options =  new MapLaunchOptions { Name = "Microsoft Building 25" };
+    var location = new Location(47.645160, -122.1306032);
+    var options = new MapLaunchOptions { Name = "Microsoft Building 25" };
 
-        try
-        {
-            await Map.OpenAsync(location, options);
-        }
-        catch (Exception ex)
-        {
-            // No map application available to open
-        }
+    try
+    {
+        await Map.OpenAsync(location, options);
+    }
+    catch (Exception ex)
+    {
+        // No map application available to open
     }
 }
 ```
 
-When opening with a `Placemark`, the following information is required:
+When using a `Placemark`, more information is required. The information helps the map app search for the place you're looking for. The following information is required:
 
 - `CountryName`
 - `AdminArea`
@@ -47,27 +44,24 @@ When opening with a `Placemark`, the following information is required:
 - `Locality`
 
 ```csharp
-public class MapTest
+public async Task NavigateToBuilding25()
 {
-    public async Task NavigateToBuilding25()
+    var placemark = new Placemark
     {
-        var placemark = new Placemark
-            {
-                CountryName = "United States",
-                AdminArea = "WA",
-                Thoroughfare = "Microsoft Building 25",
-                Locality = "Redmond"
-            };
-        var options =  new MapLaunchOptions { Name = "Microsoft Building 25" };
+        CountryName = "United States",
+        AdminArea = "WA",
+        Thoroughfare = "Microsoft Building 25",
+        Locality = "Redmond"
+    };
+    var options = new MapLaunchOptions { Name = "Microsoft Building 25" };
 
-        try
-        {
-            await Map.OpenAsync(placemark, options);
-        }
-        catch (Exception ex)
-        {
-            // No map application available to open or placemark can not be located
-        }
+    try
+    {
+        await Map.OpenAsync(placemark, options);
+    }
+    catch (Exception ex)
+    {
+        // No map application available to open or placemark can not be located
     }
 }
 ```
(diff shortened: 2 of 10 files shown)


docs/platform-integration/essentials/maps.md
```diff
@@ -1,73 +1,67 @@
 ---
-title: "Xamarin.Essentials Map"
-description: "The Map class in Microsoft.Maui.Essentials enables an application to open the installed map application to a specific location or placemark."
-ms.date: 05/26/2020
+title: "Map"
+description: "Learn how to use the .NET MAUI Map class in the Microsoft.Maui.Essentials namespace. This class enables an application to open the installed map application to a specific location or placemark."
```
davidbritch, line 3: Might be worth adding a Todo comment/issue on how to force a specific map application to open (e.g. open google maps on iOS, rather than Apple Maps). It is possible.

docs/platform-integration/essentials/media-picker.md
```diff
@@ -61,38 +72,49 @@ In your `Info.plist` add the following keys:
...
-> All methods must be called on the UI thread because permission checks and requests are automatically handled by Xamarin.Essentials.
+Each method optionally takes in a `MediaPickerOptions` parameter type that allows the `Title` to be set on some operating systems, which is displayed to the user.
 
-## General Usage
+> [!IMPORTANT]
+> All methods must be called on the UI thread because permission checks and requests are automatically handled by .NET MAUI.
+
+## Take a photo
+
+The following code sample uses the media picker to take a photo, and then calls the `LoadPhotoAsync` method to save the file:
```
davidbritch, line 107: The `LoadPhotoAsync` method saves a file? :)

docs/platform-integration/essentials/open-browser.md
```diff
@@ -36,43 +39,38 @@ Open the **AndroidManifest.xml** file under the **Properties** folder and add th
...
+    }
+    catch(Exception ex)
+    {
+        // An unexpected error occured. No browser may be installed on the device.
     }
 }
 ```
 
-This method returns after the browser was _launched_ and not necessarily _closed_ by the user.  The `bool` result indicates whether the launching was successful or not.
+This method returns after the browser is launched, not after it was closed by the user.  `Browser.OpenAsync` returns a `bool` value to indicate if the browser was successfully launched.
```
davidbritch, line 69: Maybe show it returning a value in the example above?

docs/platform-integration/essentials/includes/android-permissions.md
```diff
@@ -3,6 +3,8 @@ ms.topic: include
 ms.date: 08/04/2021
 ---
 
+### Android permissions
```
davidbritch, line 6: Just an FYI that the content in this file will definitely change (I suspect it won't be needed anymore).

docs/platform-integration/essentials/permissions.md
```diff
@@ -1,71 +1,32 @@
...
-Xamarin.Essentials attempts to abstract as many permissions as possible. However, each operating system has a different set of runtime permissions. In addition there are differences when providing a single API for some permissions. Here is a guide to the currently available permissions:
+.NET MAUI attempts to abstract as many permissions as possible. However, each operating system has a different set of permissions. Even though the API allows access to a common permission, there may be differences between operating systems related to that permission. The following table describes the available permissions:
 
 Icon Guide:
 
-* ✔️ - Supported
-* ❌ - Not supported/required
+<!-- TODO: X = not supported or required?! why required? That doesn't make sense. -->
+- ✔️ - Supported
+- ❌ - Not supported/required
```
davidbritch, line 29: Can't remember where things are up to with scrubbing UWP references, so just pointing out that a column heading in the table below says UWP.

docs/platform-integration/essentials/permissions.md
```diff
@@ -88,19 +49,67 @@ Icon Guide:
...
+```
+
+A `PermissionException` is thrown if the required permission isn't declared.
+
+> [!IMPORTANT]
+> On some platforms, a permission request can only be activated a single time. Further prompts must be handled by the developer to check if a permission is in the `Denied` state, and then ask the user to manually turn it on.
+
+## Explain why permission is needed
+
+It's best practice to explain to your user why your application needs a specific permission. On iOS, you must specify a string that is displayed to the user. Android doesn't have this ability, and also defaults permission status to `Disabled`. This limits the ability to know if the user denied the permission or if it's the first time the permission is being requested. The `ShouldShowRationale` method can be used to determine if an informative UI should be displayed. If the method returns `true`, this is because the user has denied or disabled the permission in the past. Other platforms always return `false` when calling this method.
```
davidbritch, line 100: Maybe use plural of user?

docs/platform-integration/essentials/permissions.md
```diff
@@ -123,78 +132,88 @@ public async Task<PermissionStatus> CheckAndRequestLocationPermission()
 Each permission type can have an instance of it created that the methods can be called directly.
```
davidbritch, line 132: "Each permission type can have an instance of it created so that the methods can be called directly."?

docs/platform-integration/essentials/permissions.md
```diff
@@ -123,78 +132,88 @@ public async Task<PermissionStatus> CheckAndRequestLocationPermission()
...
 
-The Permissions API was created to be flexible and extensible for applications that require additional validation or permissions that aren't included in Xamarin.Essentials. Create a new class that inherits from `BasePermission` and implement the required abstract methods.
+The Permissions API was created to be flexible and extensible for applications that require more validation or permissions that aren't included in .NET MAUI. Create a class that inherits from `Permissions.BasePermission`, and implement the required abstract methods. The following example code demonstrates the basic abstract members, but without implementation:
 
 ```csharp
-public class MyPermission : BasePermission
+public class MyPermission : Permissions.BasePermission
 {
-    // This method checks if current status of the permission
+    // This method checks if current status of the permission.
```
davidbritch, line 168: "// This method checks the current status of the permission."?

docs/platform-integration/essentials/permissions.md
```diff
@@ -123,78 +132,88 @@ public async Task<PermissionStatus> CheckAndRequestLocationPermission()
...
     public override Task<PermissionStatus> CheckStatusAsync()
     {
         throw new System.NotImplementedException();
     }
 
-    // This method is optional and a PermissionException is often thrown if a permission is not declared
+    // This method is optional and a PermissionException is often thrown if a permission is not declared.
```
davidbritch, line 174: I like that a PermissionException is "often" thrown :)

docs/platform-integration/essentials/permissions.md
```diff
@@ -123,78 +132,88 @@ public async Task<PermissionStatus> CheckAndRequestLocationPermission()
...
 
-Then you can call your new permission from Android project.
+You then check the permission in the same way as any other permission type provided by .NET MAUI:
 
 ```csharp
 await Permissions.RequestAsync<ReadWriteStoragePermission>();
 ```
 
-If you wanted to call this API from your shared code you could create an interface and use a dependency service to register and get the implementation.
+<!-- TODO: Is this still the same? Does shared code just pull in the platform specific code at compile time and thus you don't need an interface as long as you've declared the type in all platforms you're compiling for? (The rest of this section was not processed for this update and needs to be updated)-->
```
davidbritch, line 215: It's the Xamarin.Forms way of doing things. It will still work in .NET MAUI, in order to support users who are migrating their apps to MAUI. But it's the old pattern. So I recommend a todo/issue to update/replace it with the new approach (using multi-targeting with partial classes/methods).

docs/platform-integration/essentials/preferences.md
```diff
@@ -1,107 +1,141 @@
...
-var myValue = Preferences.Get("my_key", "default_value");
+string firstName = Preferences.Get("first_name", "Unknown");
+int age = Preferences.Get("age", -1);
+bool hasPets = Preferences.Get("has_pets", false);
 ```
 
-To check if a given _key_ exists in preferences:
+## Check for a key
+
+It may be useful to check if a key exists in the preferences or not. Even though `Preferences.Get` has you set a default value when the key doesn't exist, there may be cases where the key existed, but the value of the key matched the default value. So you can't rely on the default value as an indicator that the key doesn't exist. Use the `Preferences.ContainsKey` method to determine if a key exists:
```
davidbritch, line 62: Can rather than may?

davidbritch, line 62: "Even though Preferences.Get requires you...."?

docs/platform-integration/essentials/preferences.md
```diff
@@ -1,107 +1,141 @@
...
 ```csharp
 Preferences.Clear();
 ```
 
-> [!TIP]
-> The above methods take in an optional `string` parameter called `sharedName`. This parameter is used to create additional containers for preferences which are helpful in some use cases. One use case is when your application needs to share preferences across extensions or to a watch application. Please read the platform implementation specifics below.
+## Shared keys
+
+<!-- TODO: What is the difference between an extension and a different app? I'm confused (and new to .NET MAUI) -->
+<!-- TODO: Verify my statement about container is accurate -->
```
davidbritch, line 85: Verified and correct.

docs/platform-integration/essentials/preferences.md
```diff
@@ -1,107 +1,141 @@
...
 
 ```csharp
 Preferences.Clear();
 ```
 
-> [!TIP]
-> The above methods take in an optional `string` parameter called `sharedName`. This parameter is used to create additional containers for preferences which are helpful in some use cases. One use case is when your application needs to share preferences across extensions or to a watch application. Please read the platform implementation specifics below.
+## Shared keys
+
+<!-- TODO: What is the difference between an extension and a different app? I'm confused (and new to .NET MAUI) -->
```
davidbritch, line 84: This is all about iOS, specifically app extensions (https://developer.apple.com/app-extensions/) and watchOS apps (which can be linked to iOS apps, hence the need for shared preferences). I think the sentence is (largely) fine, but may need revisiting later when we find out if Essentials will still support watchOS.

docs/platform-integration/essentials/preferences.md
```diff
@@ -1,107 +1,141 @@
...
 Preferences are stored natively, which allows you to integrate your settings into the native system settings. Follow the platform documentation and samples to integrate with the platform:
 
-* Apple: [Implementing an iOS Settings Bundle](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/UserDefaults/Preferences/Preferences.html)
-* [iOS Applicaton Preferences Sample](/samples/xamarin/ios-samples/appprefs/)
-* [watchOS Settings](https://developer.xamarin.com/guides/ios/watch/working-with/settings/)
-* Android: [Getting Started with Settings Screens](https://developer.android.com/guide/topics/ui/settings.html)
-
-## Implementation Details
+- Apple: [Implementing an iOS Settings Bundle](https://developer.apple.com/library/content/documentation/Cocoa/Conceptual/UserDefaults/Preferences/Preferences.html)
+- [iOS Application Preferences Sample](/samples/xamarin/ios-samples/appprefs/)
```
davidbritch, line 103: I'd create a todo/issue to scrub for links to Xamarin samples/docs.

docs/platform-integration/essentials/preferences.md
```diff
@@ -1,107 +1,141 @@
...
 
-Uninstalling the application will cause all _Preferences_ to be removed, with the exception being apps that target and run on Android 6.0 (API level 23) or later that use [__Auto Backup__](https://developer.android.com/guide/topics/data/autobackup). This feature is on by default and preserves app data including __Shared Preferences__, which is what the **Preferences** API utilizes. You can disable this by following Google's [documentation](https://developer.android.com/guide/topics/data/autobackup).
+Uninstalling the application causes all _preferences_ to be removed, except when the app runs on Android 6.0 (API level 23) or later, while using the [**Auto Backup**`](https://developer.android.com/guide/topics/data/autobackup) feature. This feature is on by default and preserves app data, including **Shared Preferences**, which is what the **Preferences** API uses. You can disable this by following Google's [Auto Backup documentation](https://developer.android.com/guide/topics/data/autobackup).
 
 ## Limitations
 
 When storing a string, this API is intended to store small amounts of text. Performance may be subpar if you try to use it to store large amounts of text.
```
davidbritch, line 136: Not convinced that subpar will localise well. Maybe something like "Performance may be reduced if..."?

docs/platform-integration/essentials/platform-extensions.md
```diff
@@ -5,6 +5,7 @@ ms.date: 03/13/2019
 no-loc: ["Microsoft.Maui", "Microsoft.Maui.Essentials"]
 ---
 
+<!-- TODO update article -->
```
davidbritch, line 8: Definitely. Most of this has changed in .NET MAUI.

docs/platform-integration/essentials/screenshot.md
```diff
@@ -1,40 +1,39 @@
 ---
 title: "Screenshot"
-description: "Describes the Screenshot class in Microsoft.Maui.Essentials, which lets you take a capture of the current displayed screen of the app."
-ms.date: 04/14/2021
+description: "Learn how to take a screenshot of the app, using .NET MAUI. The Screenshot class in Microsoft.Maui.Essentials namespace is used to capture of the current displayed screen of the app."
```
davidbritch, line 3: "...is used to capture the..."

docs/platform-integration/essentials/screenshot.md
```diff
@@ -1,40 +1,39 @@
...
 
-    Image = ImageSource.FromStream(() => stream);
+    return ImageSource.FromStream(() => stream);
 }
 ```
 
 ## Limitations
 
-Not all views support being captured at a screen level such as an OpenGL view.
+Not all views support being captured at a screen level, such as an OpenGL view.
```
davidbritch, line 34: OpenGLView won't exist in MAUI. Not sure if this is still required (maybe it'll apply to the GraphicsView?).

davidbritch (approved):

PR #170, opened 2021-09-21
Remove WebView2 step from .NET MAUI installation.md

lutzroeder:
WebView2 is part of the Visual Studio installer.


ThisIsKirsch (approved):
Looks great to me!

Eilon (approved):

danroth27 (approved):
I think this is fine, with the assumption that you have to install VS to really use .NET MAUI. If we start pushing a CLI-only based experience, then we'll need this step again on Windows.

PR #179, opened 2021-10-06
Fix minor word transposition

ztj:
The expansion of "MAUI" was incorrect in the "under construction" warning.


davidbritch (approved):
Thanks @ztj 

Well spotted!

PR #189, opened 2021-10-25
Use media folders

davidbritch:
Re-org the live content to use media folders, rather than -images folders.


adegeo (approved):
Looks good!

PR #192, opened 2021-10-30
17.1 updates

davidbritch:
Installation updates for 17.1.


davidbritch:
This PR requires image updates, as the text now refers to an API30 emulator while the images show API28.

PR #220, opened 2021-12-06
Single project

### LINKED ISSUE #8: Single project
Ensure single project experiences are fully documented.

davidbritch:
Fixes #8


docs/fundamentals/single-project.md
```diff
@@ -1,35 +1,195 @@
...
-- App manifest
-- NuGet
-- Platform-specific code
+## Resource files
 
-All other features are being moved from their own platform-projects into platform folders in the single project.
+Resource management for cross-platform app development has traditionally been problematic. Each platform has its own approach to managing resources, that you have to understand. For example, each platform has differing image requirements that typically involves creating multiple versions of each image at different resolutions. Therefore, a single image typically has to be duplicated multiple times per platform, at different resolutions, with the resulting images having to use different filename and folder conventions on each platform.
 
-## Visual Studio changes
+.NET MAUI single project enables resource files to be stored in a single location while being consumed on each platform. This includes fonts, images, the app icon, the splash screen, raw assets, and CSS files.
```
jonathanpeppers: The mention of CSS here confuses me. Is this talking about usage with BlazorWebView? Maybe could take this one out or mention:

```suggestion
.NET MAUI single project enables resource files to be stored in a single location while being consumed on each platform. This includes fonts, images, the app icon, the splash screen, raw assets, and in the case of hybrid web applications -- CSS files.
```
(feel free to improve on my grammar here)

docs/fundamentals/single-project.md
```diff
@@ -1,35 +1,195 @@
...
 
-## Simplify development
+- A single shared project that can target Android, iOS, macOS, and Windows.
+- A simplified debug target selection for running your .NET MAUI apps.
+- Shared resource files within the single project.
+- Access to platform-specific APIs and tools when required.
+- A single cross-platform app entry point.
 
-Single project is built on top of a collection of experiences that are being simplified in .NET 6. The following list shows the experiences that will be shared in .NET MAUI single project:
+.NET MAUI single project is enabled using multi-targeting and the use of SDK-style projects in .NET 6.
```
jonathanpeppers, line 21: One thing the .NET architect's group always says is to not use the term "SDK-style". They call them "short form" MSBuild projects. As compared to the old, "long form", MSBuild projects.

If this is a common term that's used across several docs, it should be fine. This is just FYI, maybe we should check what term other MSBuild docs use?

docs/fundamentals/single-project.md
```diff
@@ -1,35 +1,195 @@
...
-  - App icons
-  - Splash screens
-  - Raw Assets
-- App manifest
-- NuGet
-- Platform-specific code
+## Resource files
 
-All other features are being moved from their own platform-projects into platform folders in the single project.
+Resource management for cross-platform app development has traditionally been problematic. Each platform has its own approach to managing resources, that you have to understand. For example, each platform has differing image requirements that typically involves creating multiple versions of each image at different resolutions. Therefore, a single image typically has to be duplicated multiple times per platform, at different resolutions, with the resulting images having to use different filename and folder conventions on each platform.
```
jonathanpeppers: The phrase, `that you have to understand`, maybe could be more general like: `to implement for each platform`. Then it would also not use "you" in the phrase.

docs/fundamentals/single-project.md
```diff
@@ -1,35 +1,195 @@
...
+```xaml
+<WebView Source="index.html" />
+```
+
+At build time, raw assets are copied to your app package.
+
+### CSS files
+
+Cascading Style Sheet (CSS) files can be added to your app project by dragging them into the _Resources_ folder of your project, and setting their build action to `MauiCss` in the **Properties** window. This creates a corresponding entry per CSS file in your project file:
+
```
jonathanpeppers, line 149: Maybe this section needs to mention Blazor or hybrid apps? Basically why a native app would need CSS.

docs/fundamentals/single-project.md
```diff
@@ -1,35 +1,195 @@
...
+In addition to this default multi-targeting approach, .NET MAUI apps can also be multi-targeted based on your own filename and folder criteria. This enables you to structure your .NET MAUI app project so that you don't have to place your platform code into sub-folders of the _Platforms_ folder. For more information, see [Configure multi-targeting](~/platform-integration/configure-multi-targeting.md).
+
+Multi-targeting can also be combined with conditional compilation so that code is targeted to specific platforms:
+
+```csharp
+#if __ANDROID__
+                  handler.NativeView.SetBackgroundColor(Colors.Red.ToNative());
+#elif __IOS__
```
jonathanpeppers: You should drop the `__` on these and just use `ANDROID` and `IOS`. These are the new defines provided by the .NET SDK, spec here:

https://github.com/dotnet/designs/blob/main/accepted/2020/net5/net5.md#vary-implementation

`__ANDROID__` and the other one still *work*, we kept them for compatibility with Xamarin apps. But I don't think any documentation should show these.

docs/fundamentals/single-project.md
davidbritch: Removed mention of CSS.

docs/fundamentals/single-project.md
davidbritch: I searched the .NET docs repo and SDK-style is a common term (e.g. https://docs.microsoft.com/en-us/dotnet/standard/frameworks). I couldn't find any mention of short form projects.

jonathanpeppers (approved):
`single-project.md` looks good 👍 . I didn't review the other pages, though, thanks.

PR #221, opened 2021-12-06
Invoking platform code

### LINKED ISSUE #196: Invoking platform code
Xamarin.Forms had the `DependencyService`. MAUI uses partial methods/classes to achieve the same result, via the magic of multi-targeting.

Doc how to do this, including:

1. Using Platform folders (all platforms vs specific platforms).
2. Using *.platform files (*.android.cs etc.).

This content should live under the Platform Integration top-level TOC node.

davidbritch:
Fixes #196


docs/platform-integration/invoke-platform-code.md
```diff
@@ -0,0 +1,144 @@
...
+1. Invoke the cross-platform API by creating an instance of the partial class and invoking its methods as required. For more information, see [Invoke the cross-platform API](#invoke-the-cross-platform-api).
+
+## Define the cross-platform API
+
+To invoke platform code from cross-platform code, the first step is to define the cross-platform API as a [partial class](/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods#partial-classes) that defines [partial method](/dotnet/csharp/programming-guide/classes-and-structs/partial-classes-and-methods#partial-classes) signatures for any operations you want to invoke on each platform.
+
+The following example shows a cross-platform API that can be used to retrieve the orientation of a device:
+
+```csharp
+namespace CallPlatformCodeDemos.Services
```
jfversluis, line 32: Not sure if there is some kind of Docs policy for it, but I know Ortinau has changed the templates to be more "C# 10" and the namespaces are file-scoped. Should we do that for the snippets/samples here to better match the templates?

docs/platform-integration/invoke-platform-code.md
```diff
@@ -0,0 +1,144 @@
...
+        public partial DeviceOrientation GetOrientation();
+    }
+}
+```
+
+The partial class is named `DeviceOrientationService`, which includes a partial method named `GetOrientation`.
+
+## Implement the API per platform
+
+After defining the cross-platform API, it must be implemented on your required platforms by defining the same partial class and the same partial method signatures, while also providing the method implementations.
```
jfversluis, line 52: Maybe emphesize that it always needs a (dummy) declaration. Even if only 1 platform, for some reason, needs an actual implementation?

docs/platform-integration/invoke-platform-code.md
```diff
@@ -0,0 +1,144 @@
...
+}
+```
+
+## Invoke the cross-platform API
+
+After providing the platform implementations, the API can be invoked from cross-platform code by creating an object instance and invoking its operation:
+
+```csharp
+using CallPlatformCodeDemos.Services;
+using Microsoft.Maui.Controls;
```
jfversluis, line 123: Following the template comment from above, the `using Microsoft.Maui.Controls` should then also not be necessary anymore

docs/platform-integration/invoke-platform-code.md
```diff
@@ -0,0 +1,144 @@
...
+...
+
+DeviceOrientationService deviceOrientationService = new DeviceOrientationService();
+DeviceOrientation orientation = deviceOrientationService.GetOrientation();
+```
+
+At build time the build system will use multi-targeting to combine the cross-platform partial class with the partial class for the target platform, and build it into the app package.
+
+## Configure multi-targeting
```
jfversluis, line 132: From this section I'm not really clear what I can use out of the box and what I need to configure if I want to use this. Maybe I need to click through to the other page, but maybe emphasize a bit more which method I can use without any additional configuration and which one is _also_ possible but requires a little work on my side.

jfversluis (approved):
Few minor things to consider but LGTM

docs/platform-integration/invoke-platform-code.md
jfversluis: I thought you could use something out of the box, but it seems you can't :D

docs/platform-integration/invoke-platform-code.md
davidbritch: Yes, I'll ensure the docs are updated for the next MAUI release.

docs/platform-integration/invoke-platform-code.md
davidbritch: Does it? I *think* this was fine when I wrote it - that MAUI just builds for the target platform, rather than all four platforms at once (which was the behaviour until recently). I'll double check.

davidbritch: I just double checked. You only need to provide a dummy/implementation on the platform you're building for. So if you only ever build for iOS/Android, no need to provide Catalyst/Windows dummies.

This was in the current public preview (with no implicit usings).

PR #225, opened 2021-12-06
Accessibility

### LINKED ISSUE #13: Semantic properties
.NET MAUI introduces semantic properties, which map plat properties to native accessibility properties. This is in addition to the existing AutomationProperties from Xamarin.Forms.

Current semantic properties:

- Description, of type string
- Hint, of type string
- HeadingLevel, of type SemanticHeadingLevel

For more info, see https://github.com/dotnet/maui/issues/469

https://devblogs.microsoft.com/xamarin/the-journey-to-accessible-apps-making-visual-text-accessible/

Additional specs:

https://github.com/dotnet/maui/issues/845
https://github.com/dotnet/maui/issues/927
https://github.com/dotnet/maui/issues/1214

Also, see the internal .NET MAUI accessibility doc.

TabIndex removed
IsTabStop removed

TabIndex and IsTabStop properties were introduced in Xamarin.Forms to help developers control the order in which UI elements would be read by a screen reader. In practice, they ended up being confusing and not meeting that need. In .NET MAUI we recommend a thoughtful design approach that orders your UI as you would want it to be read, rather than looking for programmatic ways to manipulate the structure of your interface. For cases when you must take control over the order, we recommend the community toolkit’s SemanticOrderView which will be available in the .NET MAUI version of the same.

	
	SemanticExtensions class
		SetSemanticFocus method allows you to move screen reader focus to a specific element.
		
		<VerticalStackLayout>
		    <Label
		        Text="Explore SemanticExtensions below"
		        TextColor="RoyalBlue"
		        FontAttributes="Bold"
		        FontSize="16"
		        Margin="0,10"/>
		    <Button
		        Text="Click to set semantic focus to label below"
		        FontSize="14"
		        Clicked="SetSemanticFocusButton_Clicked"/>
		    <Label
		        x:Name="semanticFocusLabel"
		        Text="Label receiving semantic focus"
		        FontSize="14"/>
		</VerticalStackLayout>
		
		private void SetSemanticFocusButton_Clicked(object sender, System.EventArgs e)
		{
		  semanticFocusLabel.SetSemanticFocus();
		}
		
		SemanticScreenReader.Announce in Essentials

cc: @rachelkang

davidbritch:
Fixes #13


docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
+---
+title: ".NET MAUI accessibility"
```
rachelkang: Based on conversations with our customers, I think it would be great if we could talk about accessibility in the context of semantics, rather than vice versa. But I'm not quite sure what the best wording would be in order to achieve this
```suggestion
title: ".NET MAUI semantics"
```
```suggestion
title: ".NET MAUI semantics for accessibility"
```

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+description: "Learn how to use the SemanticProperties class in a .NET MAUI app, so that a screen reader can speak about the user interface elements on a page."
+ms.date: 10/11/2021
+---
+
+# Accessibility
```
rachelkang:
```suggestion
# Semantics
```

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+<!-- Sample link goes here -->
+
+Accessibility is concerned with building experiences that make your apps usable by people who use technology in a wide range of environments and approach your UI with a range of needs and experiences. For some situations, accessibility requirements are imposed by law. Regardless, it's advisable to address accessibility issues regardless of legal requirements so that your apps have the largest possible audience.
```
rachelkang:
```suggestion
Semantics is concerned with building experiences that make your apps inclusive of people who use technology in a wide range of environments and approach your UI with a range of needs and experiences. In many situations, legal requirements for accessibility may provide an impetus for developers to address accessibility issues. Regardless, it's advisable to build inclusive and accessible apps so that your apps reach the largest possible audience.
```

rachelkang: If allowed, perhaps we could link my blog post here as well: https://devblogs.microsoft.com/xamarin/the-journey-to-accessible-apps/

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+Many user accessibility needs are met by assistive technology products installed by the user or by tools and settings provided by the operating system. This includes functionality such as screen readers, screen magnification, and high-contrast settings.
+
+Screen readers typically provide auditory descriptions of controls that are displayed on the screen. These descriptions help users navigate through the app and provide references to controls, such as images, that have no input or text. Screen readers are often controlled through gestures on the touchscreen, trackpad, or keyboard. For information about enabling screen readers, see [Enable screen readers](#enabling-screen-readers).
+
+Each operating system has its own screen reader with its own unique behavior and configuration. For example, most screen readers read the text associated with a control when it receives focus, enabling users to orient themselves as they navigate through the app. However, some screen readers can also read the entire app user interface when a page appears, which enables the user to receive all of the page's available informational content before attempting to navigate it.
```
rachelkang: Various screen readers can be applied to various platforms, and the platforms themselves sometimes have more than 1 screen reader offering.

```suggestion
Each operating system has its own screen readers with their own unique behavior and configuration. For example, most screen readers read the text associated with a control when it receives focus, enabling users to orient themselves as they navigate through the app. However, some screen readers can also read the entire app user interface when a page appears, which enables the user to receive all of the page's available informational content before attempting to navigate it.
```

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+.NET Multi-platform App UI (.NET MAUI) supports two approaches to providing access to the accessibility experience of the underlying native platform. *Semantic properties* are the .NET MAUI approach to providing accessibility values in apps, and are the recommended approach. *Automation properties* are the Xamarin.Forms approach to providing accessibility values in apps, and have been superseded by semantic properties. In both cases, the default accessibility order of controls is the same order in which they're listed in XAML or added to the layout. However, different layouts might have additional factors that influence accessibility order. For example, the accessibility order of `StackLayout` is also based on its orientation, and the accessibility order of `Grid` is based on its row and column arrangement.
```
rachelkang: if we can link blog posts, this would be the relevant one here: https://devblogs.microsoft.com/xamarin/the-journey-to-accessible-apps-meaningful-content-ordering/

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Semantic properties
+
+Semantic properties are used tp define information about which controls should receive accessibility focus and which text should be read aloud to the user. Semantic properties are attached properties that can be added to any element to set the underlying accessibility APIs on the native platforms.
```
rachelkang: small typo
```suggestion
Semantic properties are used to define information about which controls should receive accessibility focus and which text should be read aloud to the user. Semantic properties are attached properties that can be added to any element to set the underlying accessibility APIs on the native platforms.
```

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+In addition, the `SetValue` method can also be used to set the `SemanticProperties.Description` attached property:
+
+```csharp
+image.SetValue(SemanticProperties.DescriptionProperty, "Image of dotnet bot.");
+```
+
+> [!WARNING]
+> Don't set the `SemanticProperties.Description` attached property on a `Label`. This will stop the `Text` property being spoken by the screen reader. Instead, set the `Text` property, and the `SemanticProperties.Hint` attached property if required.
```
rachelkang: There are certain edge cases where it would be appropriate to use Description on a Label, which is why I suggest swapping out "Don't" for "Avoid". And I'm not sure if suggesting "Hint" on a label is a good idea either, so I suggest removing that
```suggestion
> Avoid setting the `SemanticProperties.Description` attached property on a `Label`. This will stop the `Text` property being spoken by the screen reader. This is because the visual text should ideally match the text read aloud by the screen reader.
```

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+```xaml
+<Label x:Name="label"
+       Text="Enter your name: " />
+<Entry SemanticProperties.Description="{Binding Source={x:Reference label} Path=Text}" />
+```
+
+Alternatively, it can be set in C# as follows:
+
+```csharp
```
rachelkang, line 78: awesome, thanks for including this example!

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+SemanticProperties.SetHint(activityIndicator, "Saving your data to the cloud.");
+```
+
+In addition, the `SetValue` method can also be used to set the `SemanticProperties.Hint` attached property:
+
+```csharp
+activityIndicator.SetValue(SemanticProperties.HintProperty, "Saving your data to the cloud.");
+```
+
+On Android, this property behaves slightly differently depending on the control it's attached to. For example, for controls without text values, such as `Switch` and `CheckBox`, the controls will display the hint with the control. However, for controls with text values, the hint is not displayed and is read after the text value.
```
rachelkang, line 111: This is weird Android stuff, so I'm wondering if it might be better to leave it out...

@PureWeen what do you think about this bit, and while you're here, the following warning as well?

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+};
+Entry entry = new Entry();
+SemanticProperties.SetDescription(entry, label.Text);
+```
+
+### Hint
+
+The `SemanticProperties.Hint` attached property represents a `string` that provides additional context to the `SemanticProperties.Description` attached property, such as the purpose of a control. Setting this property can be accomplished in XAML:
+
+```xaml
```
rachelkang, line 91: I'm having trouble determining if this is a clear example to use. I'm being extra nitpicky with this one because people often misuse Hint, and Hint often provides additional information that users can actually choose to opt out of on various platforms.

Could we perhaps change this to an example of a button with text="Send" and hint="Sends a text message"? Open to other thoughts/ideas here as well.

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+> [!WARNING]
+> The `SemanticProperties.Hint` property conflicts with the `Entry.Placeholder` property on Android, which both map to the same native property. Therefore, setting a different `SemanticProperties.Hint` value to the `Entry.Placeholder` value isn't recommended.
+
+<!-- ^ and Editor? -->
+
+### Heading levels
+
+The `SemanticProperties.HeadingLevel` attached property enables an element to be marked as a heading to organize the UI and make it easier to navigate. Some screen readers enable users to quickly jump between headings.
+
+Headings have a level from 1 to 9, and are represented by the `SemanticHeadingLevel` enumeration, which defines `None`, and `Level1` through to `Level9` members.
```
rachelkang: potential typo?
```suggestion
Headings have a level from 1 to 9, and are represented by the `SemanticHeadingLevel` enumeration, which defines `None`, and `Level1` through `Level9` members.
```

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+<StackLayout SemanticProperties.Description="Main page layout"
+             SemanticProperties.HeadingLevel="Level1">
+    ...
+</StackLayout>
+```
+
+Alternatively, it can be set in C#:
+
+```csharp
+StackLayout stackLayout = new StackLayout();
```
rachelkang: I'm not sure how it works or is supposed to work with layouts, but I think the more common use case is to set heading levels on labels! so if there are "headings" in text, large blocks of text can be skipped. (kind of like the linked "headings" in our own docs)

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+In addition, the `SetValue` method can also be used to set the `SemanticProperties.HeadingLevel` attached property:
+
+```csharp
+stackLayout.SetValue(SemanticProperties.HeadingLevelProperty, SemanticHeadingLevel.Level1);
+```
+
+<!-- Todo: is this right?
+### WebView    
+
+The native WebView control on each platform is responsible for determining which elements of its content are accessible, and each WebView control should report its HTML content as accessibility elements.
```
rachelkang: what do you mean by "each WebView control should report its HTML content as accessibility elements."?

with WebView controls, I think it's actually on the specific website / HTML content. If someone is using a WebView with a website that's not accessible, then it won't be accessible through us either, whereas if it is, then it will be.

@PureWeen do you have additional thoughts?

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+Alternatively, if the content in your `WebView` is simple, you could provide a description of the content for the `WebView` with the `SemanticProperties.Description` and `SemanticProperties.Hint` attached properties.
```
rachelkang: I'm not sure this would work, but more importantly, don't think this is advised. Did you have an example in mind? I'm not completely sure what you mean by  if the content is "simple"

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+Controls have a `SetSemanticFocus` extension method, defined in the `Microsoft.Maui` namespace, which forces screen reader focus to a specified element. For example, given a `Label` named `label`, screen reader focus can be forced to the element with the following code:
+
+```csharp
+using Microsoft.Maui;
+...
+
+label.SetSemanticFocus();
+```
+
+<!-- Todo: Any scenarios for calling UpdateSemantics or UpdateSemanticNodeInfo? -->
```
rachelkang: this was added here https://github.com/dotnet/maui/pull/1886 

I'm not sure about specific scenarios we would document, but it's mostly meant for advanced developers who are very familiar with accessibility and need to create some custom behavior

@PureWeen can add on if I'm missing something

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+## Semantic screen reader
+
+.NET MAUI Essentials includes a `SemanticScreenReader` class that enables you to instruct a screen reader to announce specified text. <!-- For more information, see [SemanticScreenReader]().-->
```
rachelkang: I think adding a couple sentences about the `Announce` method itself would be nice! Especially if we can note how it's different from `Description` since it's a method and not a property tied to a control

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+These attached properties set native accessibility values so that a screen reader can speak about the element. <!-- For more information about attached properties, see [Attached Properties](~/xamarin-forms/xaml/attached-properties.md). -->
+
+> [!IMPORTANT]
+> Using the `AutomationProperties` attached properties may impact UI Test execution on Android. The `AutomationId`, `AutomationProperties.Name` and `AutomationProperties.HelpText` properties will both set the native `ContentDescription` property, with the `AutomationProperties.Name` and `AutomationProperties.HelpText` property values taking precedence over the `AutomationId` value (if both `AutomationProperties.Name` and `AutomationProperties.HelpText` are set, the values will be concatenated). This means that any tests looking for `AutomationId` will fail if `AutomationProperties.Name` or `AutomationProperties.HelpText` are also set on the element. In this scenario, UI Tests should be altered to look for the value of `AutomationProperties.Name` or `AutomationProperties.HelpText`, or a concatenation of both.
```
rachelkang: @PureWeen I can't remember - is this fixed in maui? Is this still an issue, or can Dave remove this warning?

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+Different screen readers read different accessibility values. For example, on Windows, Narrator will prioritize `AutomationProperties.Name`, `AutomationProperties.LabeledBy`, and then `AutomationProperties.HelpText`. On Android, TalkBack may combine the `AutomationProperties.Name` and `AutomationProperties.HelpText` values. Therefore, it's recommended that thorough accessibility testing is carried out on each platform to ensure an optimal experience.
```
rachelkang: Would removing the example detract from the overall message? I'm not super confident that the priority order described is always true, but I'm also cool with leaving it if you have reason to believe it is!

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+```
+
+When this attached property is set, .NET MAUI sets the `AutomationProperties.IsInAccessibleTree` attached property to `false` on the specified element and its children.
+
+### IsInAccessibleTree
+
+> [!WARNING]
+> This attached property should typically remain unset. The majority of controls should be present in the accessibility tree, and the `AutomationProperties.ExcludedWithChildren` attached property can be set in scenarios where an element and its children need removing from the accessibility tree.
+
+The `AutomationProperties.IsInAccessibleTree` attached property, of type `bool?`, determines if the element is visible to screen readers. It must be set to `true` to use the other automation properties. This can be accomplished in XAML as follows:
```
rachelkang, line 238: > It must be set to `true` to use the other automation properties.
I can't remember if this is necessarily true. @PureWeen do you?

At least with semanticProperties, when we set a description on something, it'll automatically switch isinacessibletree=true as well (i.e. images)

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,378 @@
...
+
+The following tools can assist with your accessibility testing:
+
+- [Accessibility Insights](https://accessibilityinsights.io) for Android and Windows apps.
+- [Accessibility Scanner](https://support.google.com/accessibility/android/answer/6376570) for Android apps.
+- [Accessibility Inspector](https://developer.apple.com/library/archive/documentation/Accessibility/Conceptual/AccessibilityMacOSX/OSXAXTestingApps.html) for iOS and macOS apps.
+- [Android Studio Layout Inspector](https://developer.android.com/studio/debug/layout-inspector) for Android apps.
+- [Xcode View Debugger](https://developer.apple.com/library/archive/documentation/DeveloperTools/Conceptual/debugging_with_xcode/chapters/special_debugging_workflows.html#//apple_ref/doc/uid/TP40015022-CH9-SW2) for iOS and macOS apps.
+
+However, none of these tools can perfectly emulate the screen reader user experience, and the best way to test and troubleshoot your apps for accessibility will always be manually with screen readers.
```
rachelkang:
```suggestion
However, none of these tools can perfectly emulate the screen reader user experience, and the best way to test and troubleshoot your apps for accessibility will always be manually on physical devices with screen readers.
```

rachelkang (changes requested):
This is awesome, Dave! Thanks so much for writing all this up - it looks great.

I just had some thoughts and minor touches which I hope will be helpful. Feel free to take/leave as much as you find fitting, and let me know if you have any questions about my feedback!

And a few broader thoughts: 

- I know this is just the beginning, but I'm curious about your thoughts on creating an accessibility checklist in the future. Specifically, I think it would be nice to revamp this page: https://docs.microsoft.com/xamarin/cross-platform/app-fundamentals/accessibility

- While this is part of a longer conversation, I wanted to share that we recently learned that `native` is not an inclusive term. I'm not sure how the phasing out of such terms works on such a broad scale, but if we could substitute with `platform-specific`, `built-in`, or other appropriate alternatives wherever possible, I think that would be an awesome step in the right direction!

Thanks again for all your thorough investigating and writing as always!

docs/fundamentals/accessibility.md
davidbritch: I don't mind changing the example, but I don't see how the suggestion fits with the text about Hint providing additional context to Description. For clarity, that requires an example that sets Description and Hint. Not doing so will result in user confusion, and they will open issues. Thoughts?

rachelkang: The description is the same as the text since it's a text-based control, but I can see how that can be unclear.
How about an image button that conveys liking something? The image can be a thumbs up or heart image, description is "Like", hint is "Likes this post"

docs/fundamentals/accessibility.md
```diff
@@ -0,0 +1,395 @@
...
+For more information about Narrator, see [Complete guide to Narrator](https://support.microsoft.com/windows/complete-guide-to-narrator-e4397a0d-ef4f-b386-d8ae-c172f109bdb1).
+
+## Accessibility checklist
+
+Follow these tips to ensure that your .NET MAUI apps are accessible to the widest audience possible:
+
+<!-- markdownlint-disable MD032 -->
+
+> [!div class="checklist"]
+> - Make the user interface self-describing. Tag all the elements of your user interface with descriptive text and hints.
```
rachelkang: We only want them to use the APIs when they need to (i.e. all text-based controls should be accessible without using any semantic apis)
```suggestion
> - Make sure the user interface is self-describing. Test that all the elements of your user interface are screen reader accessible. Add descriptive text and hints when necessary.
```

docs/fundamentals/accessibility.md
davidbritch: Perfect, thanks :)

rachelkang (changes requested):
Thanks for adding the accessibility checklist! Looks great :)

I was wondering - do you think there's a way we could better tie it to the "testing accessibility" section? I found myself wanting those resources and WCAG referenced in the checklist somehow, but don't want to be redundant.

Just a thought though - do whatever you think is best!

docs/fundamentals/accessibility.md
davidbritch: I'm going to remove this WebView section, and just add a sentence somewhere saying that if a website is accessible it'll be accessible through WebView, otherwise not.

docs/fundamentals/accessibility.md
PureWeen: This warning is no longer needed. Maybe turn it into a highlight instead? Now with .NET MAUI you can use `AutomationId` and these other properties and everything will stay working!! 

We still need to wire up `HelpText/Name` inside `MAUI` but we're going to do so in such a way that it won't interfere with `AutomationId`. Currently on XF if you use the AccessibilityExperimental_Flag they also don't interfere

docs/fundamentals/accessibility.md
PureWeen: Hmmm so the main reason to use `UpdateSemanticNodeInfo` is if you're customizing code yourself through your own `AccessibilityDelegate` and you want to make sure to reapply our logic. I don't know if it's worth adding here though. That might be more useful in a section of like "Advanced" scenarios? Where we could show ways to create your own `MauiAccesssibilityDelegate` etc.. ?

docs/fundamentals/accessibility.md
PureWeen: I think also noting here that the SemanticProperty APIs perform in a much more expected and consistent way between platforms might be useful as well.

docs/fundamentals/accessibility.md
PureWeen: AFAICR it's different per platform here.  On `Android` if you set the contentDesc of an element then `Android` will just set that element as being ImportantForAccessibility and read it out, but on iOS you have to set isInAccessibleTree to `true` 

So it's probably best to just indicate that you have to set it.

docs/fundamentals/accessibility.md
davidbritch: Created https://github.com/dotnet/docs-maui/issues/236 to track this.

PR #228, opened 2021-12-06
HorizontalStackLayout and VerticalStackLayout

### LINKED ISSUE #41: VerticalStackLayout
MAUI includes a performant `VerticalStackLayout` class, that should be added to User Interface > Layouts.

Call out to use this when it fits your layout need. Only use `StackLayout` if the orientation will change at runtime (adaptive layout based on screen size or device idiom).

### LINKED ISSUE #42: HorizontalStackLayout
MAUI includes a performant `HorizontalStackLayout` class, that should be added to User Interface > Layouts.

Call out to use this when it fits your layout need. Only use `StackLayout` if the orientation will change at runtime (adaptive layout based on screen size or device idiom).

davidbritch:
Fixes #41 
Fixes #42


docs/user-interface/layouts/horizontalstacklayout.md
```diff
@@ -0,0 +1,166 @@
...
+> The `Spacing` property can be set to negative values to make child views overlap.
+
+## Position and size child views
+
+The size and position of child views within a `HorizontalStackLayout` depends upon the values of the child views' `HeightRequest` and `WidthRequest` properties, and the values of their `VerticalOptions` properties. In a `HorizontalStackLayout`, child views expand to fill the available height when their size isn't explicitly set.
+
+The `VerticalOptions` properties of a `HorizontalStackLayout`, and its child views, can be set to fields from the `LayoutOptions` struct, which encapsulates an *alignment* layout preference. This layout preference determines the position and size of a child view within its parent layout.
+
+> [!TIP]
+> Don't set the `VerticalOptions` property of a `HorizontalStackLayout` unless you need to. The default value of `LayoutOptions.Fill` allows for the best layout optimization. Changing this property has a cost and consumes memory, even when setting it back to its default value.
```
hartez: This is not the case.

docs/user-interface/layouts/horizontalstacklayout.md
```diff
@@ -0,0 +1,166 @@
...
+               VerticalOptions="Fill" />
+    </HorizontalStackLayout>
+</ContentPage>
+```
+
+In this example, alignment preferences are set on the `Label` objects to control their position within the `HorizontalStackLayout`. The `Start`, `Center`, `End`, and `Fill` fields are used to define the alignment of the `Label` objects within the parent `HorizontalStackLayout`:
+
+:::image type="content" source="media/horizontalstacklayout/alignment.png" alt-text="HorizontalStackLayout displaying aligned views screenshot.":::
+
+A `HorizontalStackLayout` only respects the alignment preferences on child views that are in the opposite direction to the orientation of the layout. Therefore, the `Label` child views within the `HorizontalStackLayout` set their `VerticalOptions` properties to one of the alignment fields:
```
hartez, line 105: Maybe "that are in the opposite direction to the orientation of the layout" -> "that are orthogonal to the orientation of the layout"?

hartez (changes requested):

docs/user-interface/layouts/horizontalstacklayout.md
davidbritch: Is it also not the case for the new StackLayout?

docs/user-interface/layouts/horizontalstacklayout.md
davidbritch: I agree that orthogonal is better, but experience tells me that (1) a lot of readers won't know what it means, (2) ESL readers won't know what it means, and (3) it won't localise well.

hartez (approved):

PR #334, opened 2022-02-28
Field modifiers clarification

### LINKED ISSUE #333: Field modifiers - clarification
There are two lines https://github.com/dotnet/docs-maui/blame/main/docs/xaml/field-modifiers.md#L18-L19 that are the same:

```md
- `internal` – specifies that the generated field for the XAML element is accessible only within types in the same assembly.
- `notpublic` – specifies that the generated field for the XAML element is accessible only within types in the same assembly.
```

Is it true that `internal` and `notpublic` actually exist? If they do, it may be better to define `notpublic` as "the same as `internal`".

davidbritch:
Fixes #333


MartyIX:
Thanks!

PR #400, opened 2022-03-24
Update app-lifecycle.md

codemv:
There was a mistake in the overridable methods for the line that reads `OnDeactivated`, which is invoked when the `OnDeactivated` event is raised. I believe that the event referred to should be the 'Deactivated' event as indicated earlier on in the article.


adegeo (approved):
Thank you for spotting and more importantly, fixing this!

PR #407, opened 2022-03-29
Inline installation instructions A|B test

davidbritch:
[Internal Preview](https://review.docs.microsoft.com/en-us/dotnet/maui/get-started/first-app?branch=dabritch-firstapp-abtest&pivots=devices-android)


ThisIsKirsch:
How are we directing people to Windows or Android? Does it happen dynamically when they land on the page or do they just go to Android by default?

davidbritch:
@ThisIsKirsch When a user first lands on that page they'll get Android (first in the switcher, due to being our default platform).

When someone switches to Windows the site will remember it and will display it automatically next time. Also, there are unique URLs for each version of the page.

lutzroeder (approved):

PR #425, opened 2022-04-08
Fix specialized page link text

joelverhagen:
`NavigationPage` and `TabbedPage` have the write link but the wrong link text.


davidbritch (approved):
Thanks @joelverhagen

PR #460, opened 2022-04-15
Update minimal iOS version for .NET MAUI Blazor apps

jfversluis:
As per [this issue](https://github.com/dotnet/maui/issues/6078) .NET MAUI Blazor on iOS has the best experience on iOS 11 and up. Reflecting this in the Docs.


davidbritch (approved):
Thanks @jfversluis

PR #494, opened 2022-05-01
Fix quotes in datatemplate.md

campersau:


adegeo (approved):
Thank you for spotting and then fixing this!

PR #558, opened 2022-05-26
Add article for some common probs

adegeo:
- Linked to wiki
- Handle these issues:
  - Templates missing
  - Can't resolve platforms after installing
  - WINDOWS #if directive
  - Contacts API can't resolve correctly on iOS and macOS

### DIFF
--- a/docs/TOC.yml
+++ b/docs/TOC.yml
@@ -5,8 +5,6 @@
       href: what-is-maui.md
     - name: Supported platforms
       href: supported-platforms.md
-    - name: Troubleshoot common problems
-      href: troubleshooting.md
     - name: Get started
       items:
         - name: Build your first app
@@ -593,3 +591,5 @@
           href: macos/deployment/overview.md
         - name: Windows
           href: windows/deployment/overview.md
+    - name: Troubleshooting
+      href: troubleshooting.md
--- a/docs/TOC.yml
+++ b/docs/TOC.yml
@@ -5,6 +5,8 @@
       href: what-is-maui.md
     - name: Supported platforms
       href: supported-platforms.md
+    - name: Troubleshoot common problems
+      href: troubleshooting.md
     - name: Get started
       items:
         - name: Build your first app
--- a/docs/troubleshooting.md
+++ b/docs/troubleshooting.md
@@ -4,7 +4,7 @@ description: "Describes the known issues and troubleshooting you can do to resol
 ms.date: 05/25/2022
 ---
 
-# Troubleshooting known issues guide
+# Troubleshooting known issues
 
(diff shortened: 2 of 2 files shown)


docs/TOC.yml
```diff
@@ -5,6 +5,8 @@
       href: what-is-maui.md
     - name: Supported platforms
       href: supported-platforms.md
+    - name: Troubleshoot common problems
```
davidbritch: Maybe just call it Troubleshooting?

davidbritch: Personally, I'd also pub this at the bottom of the TOC (experience tells me this will eventually turn into an expandable node with multiple docs under it).

docs/troubleshooting.md
```diff
@@ -0,0 +1,117 @@
+---
+title: "Troubleshoot known issues"
+description: "Describes the known issues and troubleshooting you can do to resolve these issues for a .NET Multi-platform App UI (.NET MAUI) app."
+ms.date: 05/25/2022
+---
+
+# Troubleshooting known issues guide
```
davidbritch: Do we need guide in the title?

docs/troubleshooting.md
```diff
@@ -0,0 +1,117 @@
...
+    If the result starts with `7.0`, you're running in the context of .NET 7. Use one of the fixes below.
+
+> [!TIP]
+> You can see all versions of .NET installed with the `dotnet --info` command.
+
+### Fix (1)
+
+Uninstall the .NET 7 preview.
+
+### Fix (2)
```
davidbritch, line 29: Worth stressing somewhere in this section that it's a project level fix (i.e. has to be done for every project)?

docs/troubleshooting.md
```diff
@@ -0,0 +1,117 @@
...
+
+The `WINDOWS` definition doesn't resolve correctly in the latest release of .NET MAUI. To work around this issue, add the following entry to the `<PropertyGroup>` element of your project file.
+
+```
+<DefineConstants Condition="$([MSBuild]::GetTargetPlatformIdentifier('$(TargetFramework)')) == 'windows'">$(DefineConstants);WINDOWS</DefineConstants>
+```
+
+The definitions that identify a specific version of Windows will still be missing.
+
+## ERROR Type or namespace 'Default' does not exist
```
davidbritch, line 104: Personally I don't think this is required here. It's duplication of content you've already put into the Contacts doc, and it's one of many such problems.

davidbritch (approved):
Comments inline but approving so I'm not blocking you merging it.

docs/troubleshooting.md
adegeo: I think it's good to keep. This way it bubbles up in search easier. The specific warning in the contacts article won't be unless we promote it to its own section.

PR #560, opened 2022-05-27
Convert image buttons to badge buttons

adegeo:
This changes the image buttons into true download badges

@davidbritch

[Internal preview](https://review.docs.microsoft.com/en-us/dotnet/maui/get-started/first-app?branch=pr-en-us-560)

Feel free to squash merge if it's all good.


davidbritch:
I actually hate this! I get that this approach is preferred over images. But all the buttons are different widths, and look sucky as a result. I tried putting the buttons all on one line - they won't fit.

That's why I went with images.

adegeo:
I agree on the stupid widths, but images don't scale and unfortunately, they're not translatable 😦 

image

adegeo:
Either way, stacked buttons look terrible and we should probably just clean it up with a normal link.

davidbritch (approved):
Thanks @adegeo

PR #575, opened 2022-05-31
Update navigation.md

MicrosoftButcher:
change ElephantDetailViewModel


adegeo (approved):
Verified that this PR is indeed correct.

Thank you @MicrosoftButcher for spotting this API change and submitting the PR!

PR #587, opened 2022-06-05
Adding IntentFilter Action[] constructor parameter

adenearnshaw:
IntentFilter requires an Actions array as the first parameter. Adding this, with the `ActionAppAction` declared in `Microsoft.Maui.ApplicationModel.Platform.Intent` class.


adegeo (approved):
Thanks for fixing this!

PR #595, opened 2022-06-08
Minor grammar change

martingbrown:


adegeo (changes requested):
Thank you for the contribution! According to our grammar-editing system, it's currently correct. Actually, both ways are correct. However I would prefer either:
- the emulator will run too slow.
- the emulator will run slowly. 

But I don't really care for the styling choice of
- the emulator will run too slowly.

Also, could you change line 137 to use `won't` instead of `will not`? Thanks!

martingbrown:
In hindsight I guess this isn't actually that bad. That it jarred in my head, when reading it, is probably due to the differences between American and English English.

PR #598, opened 2022-06-09
Replace macOS Catalyst app screenshot image

reddvid:
Update app screenshot on macOS to match code sample.


adegeo (approved):
This seems to match what the template has, I'll let @davidbritch confirm.

davidbritch (approved):
Thanks @reddvid

PR #636, opened 2022-06-20
fixed group name value

carpenterx:
Changed GroupName value to "colors" in order to reflect the value in the code example.


adegeo (approved):
Thank you for catching and fixing this!

PR #643, opened 2022-06-22
Update migrate.md

jamesmontemagno:


jmarti326:
LGTM!

davidbritch:
Afraid I'm closing this. The preview note is a requirement on every page until VS support for MAUI GAs. The note may be confusing, but that's because the story is confusing. But the note is accurate.

PR #648, opened 2022-06-23
Live Visual Tree

### LINKED ISSUE #378: Migrate live visual tree content

davidbritch:
Fixes #378 

[Internal Preview](https://review.docs.microsoft.com/en-us/dotnet/maui/xaml/live-visual-tree?branch=pr-en-us-648)


docs/xaml/live-visual-tree.md
```diff
@@ -0,0 +1,49 @@
...
+
+The structure of a XAML UI has a lot of elements that you're probably not directly interested in, and if you don't have a full understanding of the app's source code you might have a difficult time navigating the visual tree to find what you're looking for. Therefore, the **Live Visual Tree** window has multiple approaches that let you use the app's UI to help you find the element you want to examine:
+
+- **Select element in the running application**. You can enable this mode by clicking the **Select Element in the Running Application** button in the **Live Visual Tree** toolbar:
+
+  :::image type="content" source="media/live-visual-tree/select-element-in-running-app.png" alt-text="Screenshot of the Select Element in the Running Application button in the Live Visual Tree toolbar.":::
+
+  With this mode enabled, when you can select a UI element in the app the **Live Visual Tree** window automatically updates to show the node in the tree corresponding to that element.
+
+- **Display layout adorners in the running application**. You can enable this mode by clicking the **Display Layout Adorners in the Running Application** button in the **Live Visual Tree** toolbar:
```
davidbritch: This doesn't seem to do anything.

docs/xaml/live-visual-tree.md
```diff
@@ -0,0 +1,49 @@
...
+
+:::image type="content" source="media/live-visual-tree/just-my-xaml.png" alt-text="Screenshot of all UI elements in the Live Visual Tree window.":::
+
+Just My XAML can be disabled by selecting **Debug > Options > XAML Hot Reload** from the Visual Studio menu bar. Next, in the **Options** dialog box, ensure that **Enable Just My XAML in Live Visual Tree** is disabled:
+
+:::image type="content" source="media/live-visual-tree/just-my-xaml-disabled.png" alt-text="Screenshot of Just My XAML button disabled in the Live Visual Tree window.":::
+
+## Find a UI element
+
+The structure of a XAML UI has a lot of elements that you're probably not directly interested in, and if you don't have a full understanding of the app's source code you might have a difficult time navigating the visual tree to find what you're looking for. Therefore, the **Live Visual Tree** window has multiple approaches that let you use the app's UI to help you find the element you want to examine:
```
davidbritch: Do these three options work on all platforms?

docs/xaml/live-visual-tree.md
drasticactions: There is a bug with the current version of MAUI for WinUI where Adorners may not be drawn (https://github.com/dotnet/maui/issues/8096). It should work for Android and iOS though. If it doesn't, it needs to be filed as a bug, since it should work.

docs/xaml/live-visual-tree.md
drasticactions: Yes, they should. There should be no difference between any of them and if there are, then it's a bug that needs to be filed. The only difference is that WinUI is the only one with an in-app toolbar, but the functions from VS itself should function the same in the running application.

PR #651, opened 2022-06-24
Fix 'scollable' typo

marius-bughiu:


davidbritch (approved):
Thanks @marius-bughiu

PR #680, opened 2022-07-04
Improve grammar

loic-sharma:


adegeo (approved):
Awesome, thanks for spotting and fixing this!

PR #704, opened 2022-07-15
Fix IOS preprocessor to make code work

jfversluis:
On the .NET MAUI repo we got an [issue](https://github.com/dotnet/maui/issues/8755) stating that the code on this Docs page wasn't working. Seems that the preprocessor statement for iOS was capitalized wrong, fixing that with this PR!


davidbritch (approved):
Thanks @jfversluis

PR #711, opened 2022-07-20
Add a note for the Platforms folder

MichaelRumpler:
The files for the .csproj did not work within the Platforms folder. I lost a few weeks sulking over that and trying to find a workaround. Hopefully with that note others will place their files somewhere else from the start.

Preview does not work for the [!NOTE]. I want it to look as the [!IMPORTANT]. I just copied that and the preview for [!IMPORTANT] also doesn't work, so I guess we're good.

### DIFF
--- a/docs/platform-integration/configure-multi-targeting.md
+++ b/docs/platform-integration/configure-multi-targeting.md
@@ -73,6 +73,9 @@ This XML configures the build system to remove platform-based folder patterns un
 - Don't compile C# code that's located in the _iOS_ folder, or sub-folder of the _iOS_ folder, if you aren't building for iOS or MacCatalyst.
 - Don't compile C# code that's located in the _Windows_ folder, or sub-folder of the _Windows_ folder, if you aren't building for Windows.
 
+> [!NOTE]
+> The _Platforms_ folder gets a special handling so that files in _Platforms/iOS_ are **only** used on iOS even if you configured here that all files within any _iOS_ folder are also used for Mac Catalyst. If you want files to be used on iOS **and** Mac Catalyst, then don't place them within _Platforms_.
+
 > [!IMPORTANT]
 > Folder-based multi-targeting can be combined with filename-based multi-targeting. For more information, see [Combine filename and folder multi-targeting](#combine-filename-and-folder-multi-targeting).
 
@@ -108,3 +111,6 @@ This XML configures the build system to remove platform-based filename and folde
 - Don't compile C# code whose filename ends with _.Android.cs_, or that's located in the _Android_ folder or sub-folder of the _Android_ folder, if you aren't building for Android.
 - Don't compile C# code whose filename ends with _.iOS.cs_, or that's located in the _iOS_ folder or sub-folder of the _iOS_ folder, if you aren't building for iOS or MacCatalyst.
 - Don't compile C# code whose filename ends with _.Windows.cs_, or that's located in the _Windows_ folder or sub-folder of the _Windows_ folder, if you aren't building for Windows.
+
+> [!NOTE]
+> The _Platforms_ folder gets a special handling so that files in _Platforms/iOS_ are **only** used on iOS even if you configured here that all files within any _iOS_ folder are also used for Mac Catalyst. If you want files to be used on iOS **and** Mac Catalyst, then don't place them within _Platforms_.


adegeo:
Hi @MichaelRumpler thank you for spotting how the design has changed since this article was ported over. I really appreciate your enthusiasm, but I think this article needs much more.

I spent some time reading your original issue and scanning through this document. The problem with the note is that it's actually incorrect. You may (and I did this too) be looking at those rules for creating a way to have platforms specific files in your project, wrongly. For example:

```xml
<ItemGroup Condition="$(TargetFramework.StartsWith('net6.0-ios')) != true AND $(TargetFramework.StartsWith('net6.0-maccatalyst')) != true">
  <Compile Remove="**\iOS\**\*.cs" />
  <None Include="**\iOS\**\*.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
</ItemGroup>
```

The snippet above from the article isn't working on the principle that when targeting **-ios** or **-maccatalyst**, include these files. It's working in the reverse which excludes those files when targeting some other platform. The _Platforms_ folder is doing the same thing, it's excluding all files in the _Platforms/MacCatalyst_ folder when the platform isn't set to **-maccatalyst**. This would mean **iOS** is excluded.

In this case, the _Platforms_ folder isn't receiving any special treatment beyond what it's already configured to do, which is pointed out in the first few paragraphs in the article. I think the fix to the confusion you encountered may be in the way the example project XML is presented. Instead of making an `**/iOS/` or `*.iOS` rule, it probably should be named for all combinations. One based on `**/Apple/` for example, indicating it's both ios/maccatalyst, and also a rule for maccatalyst.

```xml
<!-- Android -->
<ItemGroup Condition="$(TargetFramework.StartsWith('net6.0-android')) != true">
  <Compile Remove="**\**\*.Android.cs" />
  <None Include="**\**\*.Android.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
</ItemGroup>

<!-- Both iOS and MacCatalyst -->
<ItemGroup Condition="$(TargetFramework.StartsWith('net6.0-ios')) != true AND $(TargetFramework.StartsWith('net6.0-maccatalyst')) != true">
  <Compile Remove="**\**\*.Apple.cs" />
  <None Include="**\**\*.Apple.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
</ItemGroup>

<!-- iOS -->
<ItemGroup Condition="$(TargetFramework.StartsWith('net6.0-ios')) != true">
  <Compile Remove="**\**\*.iOS.cs" />
  <None Include="**\**\*.iOS.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
</ItemGroup>

<!-- MacCatalyst -->
<ItemGroup Condition="$(TargetFramework.StartsWith('net6.0-maccatalyst')) != true">
  <Compile Remove="**\**\*.MacCatalyst.cs" />
  <None Include="**\**\*.MacCatalyst.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
</ItemGroup>

<!-- Windows -->
<ItemGroup Condition="$(TargetFramework.Contains('-windows')) != true ">
  <Compile Remove="**\*.Windows.cs" />
  <None Include="**\*.Windows.cs" Exclude="$(DefaultItemExcludes);$(DefaultExcludesInProjectFolder)" />
</ItemGroup>
```

Also as part of the article, I think that we should talk about how iOS isn't MacCatalyst, so you if you want to share code between just those two platforms, the `Apple` rules are the way to do that, but they shouldn't be included in the Platforms folder, well I guess really you could add the _Platforms/Apple_ folder and configure the rule for it.

What are you thoughts on this?

Regarding the NOTE/IMPORTANT, correct, you can't see those in preview here in GitHub because they're processed by the MSDocs system.

MichaelRumpler:
Yeah, if it just removes from Compile, then this explains why your Platforms rule overrules it. Good catch! As I don't know the exact Platforms rule, it was hard to see.

I'm not sure if I like having 3 folders for two platforms, but I was looking for a way to distinguish iOS from MacCatalyst. Lacking a better alternative I'd go for your new rules and change the article to that. For files with that extension it's definitely good.

adegeo:
Were you interested in making that change? Otherwise we can file a new issue to track it.

PRMerger20:
@MichaelRumpler : Thanks for your contribution! The author(s) have been notified to review your proposed change.

MichaelRumpler:
I can do it, but it would be more from you than from me with those code blocks.

adegeo:
OK. I'll close this one and follow up with a PR that handles it. Thanks again!

PR #712, opened 2022-07-22
[fix] Poor grammar in slider.md meta description

BruceMcRooster:
Fixed missing words 
“…that can be manipulated by to select…” -> “…that can be manipulated by *a user* to select…”


davidbritch (approved):
Thanks @BruceMcRooster

PR #745, opened 2022-08-05
Fixed Code snippet in ListView to have valid code

Sanguinaris:
args wasn't the name of the event (e), therefore changed that to reflect the code


davidbritch (approved):
Thanks @Sanguinaris

PR #785, opened 2022-08-12
Update cli.md

GrantMeStrength:
Added note about using Xcode-select, as this can happen if you have installed Xcode Betas or other versions (and anyone wanting to use Maui probably has) - jken@microsoft.com


davidbritch (approved):
Thanks @GrantMeStrength

PR #787, opened 2022-08-13
Update horizontalstacklayout.md

mahab339:
Corrected sentence: "In this example, the parent HorizontalStackLayout contains two nested VerticalStackLayout objects" from "In this example, the parent HorizontalStackLayout contains two nested VerticalStackLayout objects" at end of article.


davidbritch (approved):
Thanks @mahab339

PR #791, opened 2022-08-15
Typo fix - update to pair to mac article

mahab339:
Fixed typo in words order. From (If you would prefer not to install Visual Studio for 2022 Mac) to (If you would prefer not to install Visual Studio 2022 for Mac.).


TechosRX (approved):

davidbritch (approved):
Thanks @mahab339

PR #793, opened 2022-08-16
Typo fix - update to label article

mahab339:
Fixed typo of (- `TextType`, of type `TextType`, determines whether the `Label` should display plain text of HTML text.) to (... should display plain text or HTML text.).


davidbritch (approved):
Thanks @mahab339

PR #794, opened 2022-08-18
Typo fix - update to XAML compilation article

mahab339:
Correct (XAML compilation offers a number of a benefits) to (XAML compilation offers a number of benefits).


davidbritch (approved):
Thanks @mahab339

PR #796, opened 2022-08-18
Typo fix - update to FlexLayout article

mahab339:
Correct (`Wrap` property is set to to `Wrap`) to (Wrap` property is set to `Wrap`).


davidbritch (approved):
Thanks @mahab339

PR #797, opened 2022-08-18
Typo fix - update to line article

mahab339:
Correct (...and its `X2` and `Y` properties to its end point.) to (...and its `X2` and `Y2` properties to its end point.).


adegeo (approved):
Thank you for spotting and fixing this

PR #800, opened 2022-08-19
Typo fix - update to gradients article

mahab339:
Correct (fourth gradient stop specifies the color lime green at at offset of 1.0.) to (... lime green at an offset of 1.0.).


davidbritch (approved):
Thanks @mahab339

PR #801, opened 2022-08-19
Typo fix - update to linear gradient brushes article

mahab339:
Correct (`LinearGradientBrush` class also as an `IsEmpty` method that returns a `bool`) to (`LinearGradientBrush` class also has an `IsEmpty` method that returns a `bool`).


davidbritch (approved):
Thanks @mahab339

PR #803, opened 2022-08-19
Add apple to multi-targeting

adegeo:
Based on the discussion in #711

@davidbritch @MichaelRumpler


davidbritch:
I did some edits to:

- Swap Apple.cs for MaciOS.cs, which is the convention.
- Update the explanations beneath the XML.

davidbritch (approved):

PR #814, opened 2022-08-22
Typo fix - update to control scrolling in a CollectionView article

mahab339:
Change (... index of the the center item that's visible in the list.) to (... index of the center item that's visible in the list.).


davidbritch (approved):
Thanks @mahab339

PR #815, opened 2022-08-22
Typo fix - update to configure CollectionView item selection article

mahab339:
Add closing parenthese for (...SelectionChangedCommandParameter being passed to the ICommand).


davidbritch (approved):
Thanks @mahab339

PR #817, opened 2022-08-23
Replace Xamarin.Forms with .NET MAUI - update to picker item selectio…

mahab339:
…n on iOS article

Replace Xamarin.Forms with .NET MAUI (This is the default behavior in Xamarin.Forms.) sentence. Should it be omitted in case default vaule is not tested yet?


davidbritch (approved):
Thanks @mahab339

PR #822, opened 2022-08-23
Change TalkBack to VoiceOver for iOS - update to build accessible app…

mahab339:
…s with semantic properties article

Convert (**General** > **TalkBack**.) to (**General** > **Accessibility** > **VoiceOver**) in enable VoiceOver section. 
Reference: Accessibility on iOS from Xamarin.iOS docs. https://docs.microsoft.com/en-us/xamarin/ios/app-fundamentals/accessibility.


davidbritch (approved):
Thanks @mahab339

davidbritch:
@mahab339 Afraid the Xamarin.iOS docs aren't to be trusted.

I've verified that the edits I've made are correct for iOS.

PR #834, opened 2022-08-30
Acrolinx

gewarren:


docs/deployment/hot-restart.md
```diff
@@ -63,22 +63,22 @@ Perform the following steps to setup hot restart:
 
     :::image type="content" source="media/hot-restart/setup-wizard-4-add-account.png" alt-text="Screenshot of the initial fourth step in the setup hot restart wizard.":::
 
-    The **Individual account...** dialog appears.
+    The **Individual account** dialog appears.
```
gewarren, line 66: Even if it's in the menu, you're supposed to omit it in UI instructions.

docs/user-interface/handlers/create.md
```diff
@@ -2013,7 +2013,7 @@ Similarly, you can suppress the transport controls by setting the `AreTransportC
                 AreTransportControlsEnabled="False" />
 ```
 
-If you set `AutoPlay` and `AreTransportControlsEnabled` to `false`, the video won't begin playing and there'll be no way to start it playing. In this scenario you'd need to call the `Play` method from the code-behind file, or create your own transport controls.
+If you set `AutoPlay` and `AreTransportControlsEnabled` to `false`, the video won't begin playing and there will be no way to start it playing. In this scenario you'd need to call the `Play` method from the code-behind file, or create your own transport controls.
```
gewarren, line 2016: This is one contraction that Acrolinx doesn't like, because it's "rare".

adegeo (approved):

adegeo:
Thanks @gewarren !

PR #836, opened 2022-08-31
Replacing Current vs-debug-button.png

thomaselder:
Propose replacing of image in order to show consistent target device with Create a project -> Select the target device


adegeo:
Thank you for this suggestion. However, because the image cuts off the text, and the original source image (paint.net PDN file) would be out of sync with this image, I'm going to close this PR. I don't think this will trip users up as we've already received feedback of people successfully completing the tutorial. The text around that image talk about the "play" button, not the device target, so hopefully people will understand what to do.

EDIT
If I'm in here doing some work later, running through the whole tutorial again, I'll capture an updated screenshot.

PR #843, opened 2022-09-01
Update local-web-services.md

thomaselder:
Minor textual error correction


adegeo (approved):
Ha! Thanks for catching and fixing that.

PR #846, opened 2022-09-01
Typo fix - update to recognize a swipe gesture article

mahab339:
Add space between sentences of (... a swipe in more than one direction. However,...).


adegeo (approved):
Thanks again!!! You're awesome!

PR #848, opened 2022-09-01
Change tip for XAML compilation from assemblyinfo to mauiprogram

adegeo:
AssemblyInfo.cs doesn't exist as an editable file anymore, at least by default since it's auto generated. I changed it to suggest MauiProgram.cs instead.


luisquintanilla (approved):
LGTM

IEvangelist (approved):

PR #849, opened 2022-09-02
Typos fix - update to bindable properties article

mahab339:
- Correct bindable property name of "detect property changes" example explaination. From (EventName) to (IsExpanded).
- Change (which you should be handled.) to (which you should handle.).


adegeo (approved):
Thanks!

PR #865, opened 2022-09-10
Feedback on handlers

adegeo:
I reviewed the article I skipped before vacation. I added some tweaks and left a few notes to consider. We can delete the notes once you review them and decide what to do.


davidbritch (approved):
Removed the notes because one of them is incorrect, and while the other is a fair comment I already have plans to address it in another doc.

PR #868, opened 2022-09-11
Typo fix - update to preferences article

mahab339:
Remove extra backtick from (...  while using the [**Auto Backup**`]).


davidbritch (approved):
Thanks @mahab339

PR #889, opened 2022-09-18
Typo fix - update to shell navigation article

mahab339:
- Correct styling in (... returns `IReadOnlyList<Page`>, the current navigation...) to (returns `IReadOnlyList<Page>`...).
- Change type of property (Source) of class (ShellNavigationState) from (ShellNavigationState) to (ShellNavigationSource). As mentioned in subsequent explaination, and in https://github.com/dotnet/maui/blob/18ed02b41a35edad98aef79ec677c42ebff52890/src/Controls/src /Core/Shell/ShellNavigatedEventArgs.cs


davidbritch (approved):
Thanks @mahab339

PR #890, opened 2022-09-18
Typo fix - update to target multiple platforms article

mahab339:
Change (... and code and that starts the app on each platform...) to (... and code that starts... ).


davidbritch (approved):
Thanks @mahab339

PR #910, opened 2022-09-22
Typo fix - update to define an EmptyView for a CarouselView article

mahab339:
Remove extra open square bracket from (... root element of the [`EmptyView`. ).


davidbritch (approved):
Thanks @mahab339

PR #911, opened 2022-09-22
Typo fix - update to configure CarouselView interaction article

mahab339:
Change (... displayed item CarouselView can be programmatically...) to (... displayed item in a CarouselView...).


davidbritch (approved):
Thanks @mahab339

PR #915, opened 2022-09-23
Typo fix - update to triggers article

mahab339:
Change (... provide a default value for the target property (eg. Text="">... ) to (... (eg. Text="")... ).


davidbritch (approved):
Thanks @mahab339

PR #916, opened 2022-09-23
Typo fix - update to attached properties article

mahab339:
Change styles of setpropertyname and getpropertyname occurences from (`Set`*PropertyName*) and (`Get`*PropertyName*) to (`SetPropertyName`) and (`GetPropertyName`). Please ignore if they are not typo and is intended. Thanks.


davidbritch:
This was intended. `GetPropertyName` means there's a member called `GetPropertyName`, whereas the member would be `Get` followed by the property name e.g. `GetAutoPlay`.

PR #957, opened 2022-10-03
Typo fix - update to create a custom control using handlers article

mahab339:
- Fix case of (ReplaceCurrentItemWIthPlayerItem).
- Change (These requirements imply that that video...) to (These requirements imply that video...).


davidbritch (approved):
Thanks @mahab339

PR #975, opened 2022-10-10
Fix broken link on dotnet docs updates text in whats-new page

wmundev:
Fixes broken link on dotnet docs updates text in the whats-new page that i found


davidbritch (approved):
Thanks @wmundev

PR #994, opened 2022-10-12
Typo fix - update to GraphicsView article

mahab339:
Change (`Center`, of type `PointF`. which specifies...) to (...of type `PointF`, which...).


davidbritch (approved):
Thanks @mahab339

PR #1005, opened 2022-10-15
Typo fix - update to geometries article

mahab339:
- Change `SweepDirection` enumeration members description to reflect corresponding values.
- Change  (... type `GeometryCollection`, which species the objects that define...) to (.. type `GeometryCollection`, which specifies...).


davidbritch (approved):
Thanks @mahab339

PR #1006, opened 2022-10-15
Fix for error produced from example - update to image article

mahab339:
Change (CacheValidity="10:00:00.0") to (CacheValidity="10:00:00"). 
Reason: extra zero produces error:  "The TimeSpan could not be parsed because at least one of the numeric components is out of range or contains too many digits.". 
Project workload: .NET Maui 7.0.0.


davidbritch (approved):
Thanks @mahab339

PR #1007, opened 2022-10-16
Typo fix - update to path markup syntax article

mahab339:
- Change (`-Infinity` represents `double.NegativeInfinity`.) to (`-Infinity` represents...). 
- Change (... two specified control point.) to (...  two specified control points.)


davidbritch (approved):
Thanks @mahab339

PR #1009, opened 2022-10-17
Geolocation - add warning about null return value

### LINKED ISSUE #10562: Documentation for Geolocation returning null missing
### Description

https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/device/geolocation?tabs=android

`Geolocation.Default.GetLocationAsync()` sometimes returns `null`.
There is no comment on why and how to handle this.
It sometimes returns null when device is locked (Activity is Stopped).
It sometimes returns null from the start of the GPS check and never recovers (GPS permission is granted, GPS is turned on, GPS in the same location normally returns a value).
In other cases, when device is locked (Activity is Stopped), it returns some kind of cached GPS.

What does `null` returned by `GetLocationAsync()` mean?
Does it mean that GPS location could not be retrieved within the timeout set in `GeolocationRequest` `CancellationToken`?

Thank you for adding documentation

### Steps to Reproduce

call `Geolocation.Default.GetLocationAsync()` and receive `null`

### Link to public reproduction project repository

none

### Version with bug

6.0.486 (current)

### Last version that worked well

Unknown/Other

### Affected platforms

Android

### Affected platform versions

Android 8, Android 12

### Did you find any workaround?

_No response_

### Relevant log output

_No response_

davidbritch:
Fixes https://github.com/dotnet/maui/issues/10562

### DIFF
--- a/docs/platform-integration/device/geolocation.md
+++ b/docs/platform-integration/device/geolocation.md
@@ -137,6 +137,9 @@ The following code example demonstrates how to request the device's location, wh
 
 Not all location values may be available, depending on the device. For example, the `Altitude` property might be `null`, have a value of 0, or have a positive value indicating the meters above sea level. Other values that may not be present include `Speed` and `Course`.
 
+> [!WARNING]
+> `GetLocationAsync` can return `null` in some scenarios. This indicates that the underlying platform is unable to obtain the current location.
+
 ## Accuracy
 
 The following sections outline the location accuracy distance, per platform:


janseris:
>GetLocationAsync can return null in some scenarios. This indicates that the underlying platform is unable to obtain the current location.

That is not very helpful. For a robust app, it must be clear why the platform returns null and how to assure it returns real values again.
I am particularly interested in Android platform.

davidbritch:
@janseris Google's Android docs and Apple's iOS docs don't make clear in which circumstances getting the current location returns null.

janseris:
> @janseris Google's Android docs and Apple's iOS docs don't make clear in which circumstances getting the current location returns null.

Thanks. If it is like this it would be very nice if the info would be there that it's a direct call to Android API and thus it behaves like Android API and the programmer should be looking for issues/questions on Android API for more details.

PR #1014, opened 2022-10-20
Update image.md

tiwahu:
Fix XAML example of a 10-day TimeSpan.


davidbritch (approved):
Thanks for the fix @tiwahu

PR #1046, opened 2022-10-29
Double quotes missing in blazorwebview.md

joel-heath:
Added some missing double quotes in the ContentPage tag in docs/user-interface/blazorwebview.md


davidbritch (approved):
Thanks @joelheath24

PR #1069, opened 2022-11-10
Typo fix - update to paint graphical objects article

mahab339:
Change (... stops can be added to a `GradientPoint` object...) to (... `GradientPaint`...).


davidbritch (approved):
Thanks @mahab339

PR #1078, opened 2022-11-17
Typo fix - update to open the browser article

mahab339:
Change (...., add the following `queries/intent` nodes the `manifest` node:...) to (... nodes inside of the `manifest` node:...).


davidbritch (approved):
Thanks @mahab339

PR #1084, opened 2022-11-20
Typo fixes - update to email article

mahab339:
- Change (..., add the following `queries/intent` nodes the `manifest` node:...) to (...`queries/intent` nodes in the `manifest`...).
- Update code example description from (The following example demonstrates adding arbitrary text to a file, and then adding it to the email.) to (... demonstrates adding an image file to the email attachments.).
- Add header (## Platform Differences).
- Change (Only supports `EmailBodyFormat.PlainText` as the EmailBodyFormat.`BodyFormat`.) to (... as the `EmailMessage.BodyFormat`.).


davidbritch (approved):
Thanks @mahab339

PR #1085, opened 2022-11-20
Typo fix - update to phone dialer article

mahab339:
Change (..., add the following queries/intent nodes the manifest node:...) to (...queries/intent nodes in the manifest...).


davidbritch (approved):
Thanks @mahab339

PR #1113, opened 2022-12-06
Update flyout.md

AndreasReitberger:
The 
```cs
<VisualState x:Name="Normal">
                            <VisualState.Setters>
                                <Setter Property="BackgroundColor" Value="Transparent" />
                            </VisualState.Setters>
                        </VisualState>
```
needs to be wrapped in `<VisualState.Setters>` tag, otherwise VS shows an error and cannot build.


davidbritch (approved):
Thanks @AndreasReitberger

PR #1227, opened 2023-02-01
Add caution about iOS AppIcon background color

glihm:
Currently, App Store Connects upload systematically fails when a `<MauiIcon>` is added without any `Color` value. By default, the background is then considered as transparent, which triggers a 100% fail while uploading the archive to App Store Connect (even by using Xcode or command line tools provided by Apple). In order to fix that, related to this [issue](https://github.com/dotnet/maui/issues/11124) in the `maui` repository, it can be very helpful to warn users that a background must be provided. I tested on `svg` and `png` format, this solved the issue as mentioned in the linked issue.

I don't think this is an issue on it's own, it's the behavior of the framework. However, Apple is enforcing a non-transparent background.

For this reason, applying a **CAUTION** is, in my opinion, relevant. We are saving lot's of time to developers and avoiding tremendous frustration. Or perhaps a `WARN` may be good enough to catch user attention while reading iOS tab. But something must be told! :)


davidbritch (approved):
Thanks @glihm - much appreciated!

glihm:
Thank you for the rewording and such reactivity!

PR #1240, opened 2023-02-03
Ensure Rosetta is installed on Apple Silicon Macs

ghost:
User must accept Rosetta install prompt when installing Mono/Visual Studio

### DIFF
--- a/docs/ios/pair-to-mac.md
+++ b/docs/ios/pair-to-mac.md
@@ -20,7 +20,7 @@ Pair to Mac enables the following software development workflow:
 - It's possible to use the Windows command-line to build iOS apps.
 
 > [!NOTE]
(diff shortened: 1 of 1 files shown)


MartyIX:
Have you tried building your MAUI app for arm64? By default on Apple Silicon, you'll still compile for x64 (and then I expect that that's the reason you need Rosetta).

ghost:
> 

The error does not only raise on build phase. The problem is that if you have not Rosetta installed/enabled it will raise this exception: https://developercommunity.visualstudio.com/t/bad-cpu-type-when-pairing-to-mac/1478748 when pairing with Mac.

Another scenario is if Mono is not installed, and VS will ask to install it. But this automated behavior won't check for Rosetta, and Mac pairing will be interrupted with the error above.

davidbritch (approved):
Many thanks @gbarceloPIB

PR #1263, opened 2023-02-12
Replace method name `Load` with `Delete`

sd0e:
Hey, I was reading the MVVM tutorial and on the page for [creating the Note viewmodel](https://learn.microsoft.com/en-us/dotnet/maui/tutorials/notes-mvvm/?view=net-maui-7.0&tutorial-step=5) I noticed that it says to add the `Save` and `Load` methods, but provided the implementation for `Save` and `Delete` methods. I think this may be a mistake? If not please correct me! Thank you


davidbritch (approved):
Many thanks @sd0e

PR #1278, opened 2023-02-21
Fixed .NET Upgrade Assistant installation instruction

AndreKraemer:
There was a copy/paste error in the installation section for .NET Upgrade Assistant. The installation instructions said that the tool should be installed with the following command:

`dotnet tool update -g upgrade-assistant`

In order to install the tool you have to use `install` instead of `update`. So the correct command is:

`dotnet tool install -g upgrade-assistant`


adegeo (approved):

adegeo:
Thanks!

PR #1301, opened 2023-02-24
Update app-lifecycle.md

MartyIX:


davidbritch:
Hi @MartyIX 

For a user who copy/pastes this code they'd either need to have enabled nullables in their .csproj or add a #nullable enable to the MauiProgram class, right?

MartyIX:
That's correct. Isn't it the default for new MAUI apps?

davidbritch:
Afraid not. See https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj

MartyIX:
Then I guess I should close this PR.

PR #1321, opened 2023-03-01
Add code highlight for AllNotesPage.xaml snippet

mdelez:
If a user is following this guide after following the [previous guide](https://learn.microsoft.com/en-us/dotnet/maui/tutorials/notes-app/?view=net-maui-7.0) and isn't just pasting the code, the binding context for the ItemsSource attribute is also updated. 

Another issue is that the code highlights are one line off. I'm not sure why this is but this is the only .xaml file in this directory that has comments so maybe that has something to do with it.


mdelez:
> 

@dotnet-policy-service agree

davidbritch (approved):
Thanks @mdelez

PR #1336, opened 2023-03-07
Android manifest

### LINKED ISSUE #1279: Android Manifest
The Xamarin.Android docs have [this doc](https://learn.microsoft.com/en-us/xamarin/android/platform/android-manifest) about Android manifests. It needs bringing into the MAUI docset, with updates where required.

This content is required by the product team, as the Android Manifest editor in VS will be linking to it.

Also, mention the new editor (only temporarily until there's a manifest editor doc).

---
[Associated WorkItem - 66541](https://dev.azure.com/msft-skilling/Content/_workitems/edit/66541)

davidbritch:
[Internal Preview](https://review.learn.microsoft.com/en-us/dotnet/maui/android/manifest?view=net-maui-7.0&branch=pr-en-us-1336)

Fixes #1279

### DIFF
--- a/docs/android/manifest.md
+++ b/docs/android/manifest.md
@@ -8,7 +8,7 @@ ms.date: 03/07/2023
 
 Every .NET Multi-platform App UI (.NET MAUI) app on Android has an *AndroidManifest.xml* file, located in the *Platforms\\Android* folder, that describes essential information about your app to build tools, the Android operating system, and Google Play.
 
-The manifest file for your .NET MAUI Android app is generated as part of the .NET MAUI build process on Android. This build process takes the XML in the *Platforms\\Android\\AndroidManifest.xml* file, and merges it with any XML that is generated from specific attributes that are added to your classes. The resulting manifest file can be found in the *obj* folder. For example, it can be found at *obj\\Debug\\net7.0-android\\AndroidManifest.xml* for debug builds on .NET 7.
+The manifest file for your .NET MAUI Android app is generated as part of the .NET MAUI build process on Android. This build process takes the XML in the *Platforms\\Android\\AndroidManifest.xml* file, and merges it with any XML that's generated from specific attributes on your classes. The resulting manifest file can be found in the *obj* folder. For example, it can be found at *obj\\Debug\\net7.0-android\\AndroidManifest.xml* for debug builds on .NET 7.
 
 <!-- TODO: Uncomment when VS 17.6 P2 is released
 > [!NOTE]
@@ -19,7 +19,7 @@ The manifest file for your .NET MAUI Android app is generated as part of the .NE
 
 All .NET MAUI app's have a `MainActivity` class that derives from <xref:Android.App.Activity>, via the `MauiAppCompatActivity` class, and that has the <xref:Android.App.ActivityAttribute> applied to it. Some apps may include additional classes that derive from <xref:Android.App.Activity> and that have the <xref:Android.App.ActivityAttribute> applied.
 
-At build time, assemblies are scanned for non-`abstract` classes that derive from <xref:Android.App.Activity> and have the <xref:Android.App.ActivityAttribute> declared on them. These classes and attributes are used to build the manifest. For example, consider the following example:
+At build time, assemblies are scanned for non-`abstract` classes that derive from <xref:Android.App.Activity> and that have the <xref:Android.App.ActivityAttribute> applied. These classes and attributes are used to generate the app's manifest. For example, consider the following code:
 
 ```csharp
 using Android.App;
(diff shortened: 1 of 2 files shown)


docs/android/manifest.md
```diff
@@ -0,0 +1,246 @@
...
+This example causes the following XML fragment to be added to the manifest file:
+
+```xml
+<activity android:name="crc64bdb9c38958c20c7c.MainActivity" />
+```
+
+> [!NOTE]
+> <xref:Android.App.ActivityAttribute> has no effect on `abstract` types.
+
+## Activity name
```
davidbritch, line 56: Do we really need this section?

docs/android/manifest.md
```diff
@@ -0,0 +1,246 @@
...
+This example produces the following XML fragment:
+
+```xml
+<activity android:name="companyname.mymauiapp.activity" />
+```
+
+> [!NOTE]
+> You should only use the `Name` property for backward-compatibility reasons, as such renaming can slow down type lookup at runtime.
+
+## Activity title bar
```
davidbritch: Is this a likely scenario for MAUI?

davidbritch:
@jonathanpeppers I was asked to bring [this doc](https://learn.microsoft.com/en-us/xamarin/android/platform/android-manifest) about Android manifests into this repo, and update for MAUI. This content is required by the product team, as the Android Manifest editor in VS will be linking to it.

Can you take a quick look to see if it's accurate for MAUI?

docs/android/manifest.md
jonathanpeppers: This is how you could get a readable Java name for your activity, it can be useful if another Android app needs to be able to open your app (like Facebook login, for example). Or if you simply have a script for launching your app & testing startup time, etc.

docs/android/manifest.md
jonathanpeppers: This one could mention `$(ApplicationTitle)` instead:

https://github.com/dotnet/maui/blob/73098ab6473eb191d55ed991c474ab68d0cf493d/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj#L15

We might have another "single project" doc we could link to.

docs/android/manifest.md
```diff
@@ -0,0 +1,246 @@
...
+
+## Permissions
+
+When you add permissions to an Android app, they're recorded in the manifest file. For example, if you set the `ACCESS_NETWORK_STATE` permission, the following element is added to the manifest file:
+
+```xml
+<uses-permission android:name="android.permission.ACCESS_NETWORK_STATE" />
+```
+
+Debug builds automatically set the `INTERNET` permission. However, this permission is set in the generated *obj\\Debug\\net7.0-android\\AndroidManifest.xml* file. If you examine the release build version of the manifest at *obj\\Release\\net7.0-android\\AndroidManifest.xml*, the `INTERNET` permission is not present.
```
jonathanpeppers: This is true, except we put it in the MAUI project template:

https://github.com/dotnet/maui/blob/73098ab6473eb191d55ed991c474ab68d0cf493d/src/Templates/src/templates/maui-mobile/Platforms/Android/AndroidManifest.xml#L5

(Just because nearly every app uses the internet)

docs/android/manifest.md
```diff
@@ -0,0 +1,246 @@
...
+
+For example, the following `Application` attribute could be added to *MainApplication.cs* to indicate that the app can be debugged, that its user-readable name is "My Maui App", and that it uses the `Maui.SplashTheme` style as the default theme for all activities:
+
+```csharp
+using Android.App;
+using Android.Runtime;
+
+namespace MyMauiApp;
+
+[Application(Debuggable = true, Label = "My Maui App", Theme = "@style/Maui.SplashTheme")]
```
jonathanpeppers: I don't think we want an example of adding `Debuggable=true` here. Debug builds will do this for you, and Release builds automatically turn it off. What it does is let things like debuggers and other developer tooling attach to your app -- to do things like inspect memory, etc.

I think it is a security risk to submit an app to Google Play with this setting -- they might even block your submission.

docs/android/manifest.md
```diff
@@ -0,0 +1,246 @@
...
+      protected override MauiApp CreateMauiApp() => MauiProgram.CreateMauiApp();
+}
+
+```
+
+This declaration causes the following XML fragment to be generated in *obj\\Debug\\net7.0-android\\AndroidManifest.xml*:
+
+```xml
+<application android:label="MyMauiApp" android:debuggable="true" android:theme="@style/Maui.SplashTheme" ...>
+```
```
jonathanpeppers: We could still give this example, but maybe we just say Debug apps do this and Release apps don't.

jonathanpeppers (commented):
Overall this doc looks pretty relevant, I think it would be helpful to MAUI developers that want to understand the Android manifest further and do Android-specific things.

PR #1345, opened 2023-03-09
run quest nightly

BillWagner:
Quest runs nightly, rather than on each label update.

The following changes make that happen:

- The bulk action runs once a day, at a hopefully convenient time.
- The single item workflow runs only in response to a workflow dispatch event (user started)
- For security reasons, limit the permissions to write *issues*, not *contents*

### DIFF
--- a/.github/workflows/quest-bulk.yml
+++ b/.github/workflows/quest-bulk.yml
@@ -1,5 +1,7 @@
 name: "bulk quest import"
 on:
+  schedule:
+    - cron: '0 2 * * *' # UTC time, that's 9:00 pm EST, 6:00 pm PST.
   workflow_dispatch:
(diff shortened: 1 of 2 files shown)


.github/workflows/quest-bulk.yml
```diff
@@ -1,5 +1,7 @@
 name: "bulk quest import"
 on:
+  schedule:
+    - cron: '0 2 * * *' # UTC time, that's 9:00 pm EST, 6:00 pm PST.
```
BillWagner, line 4: @davidbritch I picked a different time to match likely downtime in the UK. Make a suggestion if you want a different time.

davidbritch (approved):
All good with me!

PR #1372, opened 2023-03-28
Only run quest in the dotnet org

BillWagner:
It fails when it runs in forks, so only run in the dotnet org.


davidbritch (approved):
LGTM

PR #1502, opened 2023-05-22
Android linking

### LINKED ISSUE #1390: Android linking
Now that Android publishing docs are complete, add a doc on the Android linker. The Xamarin.Androd doc on linking can be found at https://learn.microsoft.com/en-us/xamarin/android/deploy-test/linker but is largely irrelevant for MAUI.

By default, trimming is disabled for debug builds and enabled for release builds, and then the standard trimming techniques for controlling preservation are used (see https://learn.microsoft.com/en-us/dotnet/maui/ios/linking?tabs=vs#preserve-code for iOS). Note that the `Preserve` attribute, from Xamarin, isn't supported for MAUI.

VSMac still shows the old Xamarin UI for linking. However, AndroidLinkMode=None or AndoidLinkMode=SdkOnly will still work due to redirecting to the PublishTrimmed property (AndroidLinkMode=None same as PublishTrimmed=false, AndroidLinkMode=SdkOnly same as PublishTrimmed=true).

---
[Associated WorkItem - 91422](https://dev.azure.com/msft-skilling/Content/_workitems/edit/91422)

davidbritch:
Fixes #1390 

Also now shares the "Preserve code" content with iOS/MacCat via an include file.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/android/deployment/publish-cli.md](https://github.com/dotnet/docs-maui/blob/5c44a42ef27cdcf15e04a0fd151a671687257fb6/docs/android/deployment/publish-cli.md) | [Publish an Android app using the command line](https://review.learn.microsoft.com/en-us/dotnet/maui/android/deployment/publish-cli?branch=pr-en-us-1502) |
| [docs/android/linking.md](https://github.com/dotnet/docs-maui/blob/5c44a42ef27cdcf15e04a0fd151a671687257fb6/docs/android/linking.md) | [docs/android/linking](https://review.learn.microsoft.com/en-us/dotnet/maui/android/linking?branch=pr-en-us-1502) |
| [docs/ios/linking.md](https://github.com/dotnet/docs-maui/blob/5c44a42ef27cdcf15e04a0fd151a671687257fb6/docs/ios/linking.md) | [Linking a .NET MAUI iOS app](https://review.learn.microsoft.com/en-us/dotnet/maui/ios/linking?branch=pr-en-us-1502) |
| [docs/mac-catalyst/linking.md](https://github.com/dotnet/docs-maui/blob/5c44a42ef27cdcf15e04a0fd151a671687257fb6/docs/mac-catalyst/linking.md) | ["Linking a .NET MAUI Mac Catalyst app"](https://review.learn.microsoft.com/en-us/dotnet/maui/mac-catalyst/linking?branch=pr-en-us-1502) |


docs/android/linking.md
```diff
@@ -0,0 +1,155 @@
...
+> Visual Studio for Mac sets the `$(AndroidLinkMode)` MSBuild property in your app's project file. `<AndroidLinkMode>None</AndroidLinkMode>` maps to `<PublishTrimmed>false</PublishTrimmed>` and `<AndroidLinkMode>SdkOnly</AndroidLinkMode>` maps to `<PublishTrimmed>true</PublishTrimmed>`.
+
+----
+
+## Preserve code
+
+When you use the linker, it sometimes remove code that you might have called dynamically, even indirectly. You can instruct the linker to preserve members by annotating them with the [`DynamicDependency`](xref:System.Diagnostics.CodeAnalysis.DynamicDependencyAttribute) attribute. This attribute can be used to express a dependency on either a type and subset of members, or at specific members.
+
+> [!IMPORTANT]
+> Every member that isn't statically linked by the app is subject to be removed.
```
jonathanpeppers: Using the phrase "statically linked" sounds like a C/C++ term to me.

Maybe we could phrase it as:

```suggestion
> Every member in the BCL that cannot be statically determined to be used by the app is subject to be removed.
```
That still sounds wordy to me... But maybe something like that?

docs/android/linking.md
```diff
@@ -0,0 +1,155 @@
...
+[DynamicDependency("GenericMethodName``1")]
+[DynamicDependency("GenericMethod``2(``0,``1)")]
+[DynamicDependency("MethodWithGenericParameterTypes(System.Collections.Generic.List{System.String})")]
+[DynamicDependency("MethodOnGenericType(`0)", "GenericType`1", "UnreferencedAssembly")]
+[DynamicDependency("MethodOnGenericType(`0)", typeof(GenericType<>))]
+```
+
+## Preserve assemblies
+
+It's possible to specify assemblies that should be excluded from the linking process, while allowing other assemblies to be linked.
```
jonathanpeppers: You could add a comment to say why someone would use the XML over attributes:
```
This is useful as a last resort, when you can't easily use `[DependencyAttribute]` or do not control the code that is being linked away.
```

docs/android/linking.md
```diff
@@ -0,0 +1,155 @@
...
+If you have a library in your project, or you're a developer of a reusable library and you want the linker to treat your assembly as linkable, you can mark the assembly as linker safe with the [`AssemblyMetadata`](xref:System.Reflection.AssemblyMetadataAttribute) attribute:
+
+```csharp
+[assembly: AssemblyMetadata("IsTrimmable", "True")]
+```
+
+Alternatively, you can set `<IsTrimmable>true</IsTrimmable>` in a `<PropertyGroup>` tag in the project file for the assembly. This marks your assembly as "trimmable" and enable trim warnings for that project. Being "trimmable" means your assembly is considered compatible with trimming and should have no trim warnings when the assembly is built. When used in a trimmed app, the assembly's unused members will be removed in the final output.
+
+> [!NOTE]
+> If the `IsTrimmable` MSBuild property is set for an assembly, this overrides the `IsTrimmable` attribute. This enables you to opt an assembly into trimming even if it doesn't have the attribute, or to disable trimming of an assembly that has the attribute.
```
jonathanpeppers: The MSBuild property basically just puts this in the assembly for you:
```csharp
[assembly: AssemblyMetadata("IsTrimmable", "True")]
```

So maybe could rephrase this to mention it?

docs/android/linking.md
```diff
@@ -0,0 +1,155 @@
...
+```
+
+In this example, the XML file specifies a method that's dynamically accessed by the app, which is excluded from linking.
+
+When an assembly, type, or member is listed in the XML, the default action is preservation, which means that regardless of whether the linker thinks it's used or not, it's preserved in the output.
+
+> [!NOTE]
+> The preservation tags are ambiguously inclusive. If you don’t provide the next level of detail, it will include all the children. If an assembly is listed without any types, then all the assembly’s types and members will be preserved.
+
+## Mark an assembly as linker safe
```
jonathanpeppers: Should we have a general `[!NOTE]` section saying you should really know what you're doing if doing this? So far only Microsoft internal folks and maybe [Uno](https://platform.uno/) have done this successfully.

Library authors would probably want to enable these settings;
```xml
<SuppressTrimAnalysisWarnings>false</SuppressTrimAnalysisWarnings>
<TrimmerSingleWarn>false</TrimmerSingleWarn>
```
And then solve all the warnings that result.

I don't know if we have any existing docs on doing this. But this issue talks about it a bit:

https://github.com/xamarin/xamarin-android/issues/5652

jonathanpeppers (approved):

PR #1507, opened 2023-05-24
Mono interpreter

### LINKED ISSUE #1438: Mono interpreter
Many of the iOS-specific issues encountered in .NET MAUI are down to use of unsupported aspects of generics due to AOT compilation. In these circumstances, the Mono interpreter should be enabled (`<UseInterpreter>true</UseInterpreter>`). A benefit of this is reduction in overall IPA size.

The Mono interpreter is safe to enable for production apps. However, there are some drawbacks:

- App size may be impacted for better or worse (it can go either way, it depends on the app code/usage).
- Execution speed will be slower. However, this may range from ‘unmeasurable’ to ‘unacceptable’. You many not notice any performance hit at all for LOB apps but math-heavy apps/games could be a different story. You’d have to measure and assess if you’re ok with the performance hit.
- Crash reports are less helpful since they’ll not show the real frames, but instead mention the interpreter executing things.

Use of the interpreter doesn’t have to be an all or nothing proposition. You can enable the interpreter on a more granular per-assembly basis using the `MtouchInterpreter` property. This enables you to:
 
- Interpret all assemblies by specifying all or AOT compile everything by specifying -all
- Interpret individual assemblies by specifying TheAssembly.dll or AOT compile individual assemblies by specifying -TheAssembly.dll
- Mix and match to interpret some assemblies and AOT compile other assemblies

For example, in the app .csproj you can use an additive approach to AOT compile everything except for a specific third-party component for iOS release builds:

```xml
<PropertyGroup Condition=" '$(Configuration)|$(RuntimeIdentifier)'=='Release|maccatalyst-arm64' " >
    <MtouchInterpreter>-all,ThirdPartyComponent.dll</MtouchInterpreter>
</PropertyGroup>
```

The value is a comma-separated list of assemblies to interpret (if prefixed with a minus sign, the assemblies will be AOT-compiled instead), and ‘all’ can be used to specify all assemblies.
 
This means that you can either start with interpreting everything and then selecting which assemblies to AOT compile, or the opposite: AOT everything, and then select which assemblies to interpret:

```xml
<PropertyGroup>
    <!-- Interpret everything, except System.Xml.dll -->
    <MtouchInterpreter>all,-System.Xml.dll</MtouchInterpreter>
    <!-- AOT everything, except System.Numerics.dll, which will be interpreted -->
    <MtouchInterpreter>-all,System.Numerics.dll</MtouchInterpreter>
</PropertyGroup>
```

See also: https://github.com/dotnet/maui/issues/13019

---
[Associated WorkItem - 92650](https://dev.azure.com/msft-skilling/Content/_workitems/edit/92650)

davidbritch:
Fixes #1438 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/macios/interpreter.md](https://github.com/dotnet/docs-maui/blob/fcbd7820c16525a5bb2fe5590491f7d8d1b031f0/docs/macios/interpreter.md) | [docs/macios/interpreter](https://review.learn.microsoft.com/en-us/dotnet/maui/macios/interpreter?branch=pr-en-us-1507) |

### DIFF
--- a/docs/macios/interpreter.md
+++ b/docs/macios/interpreter.md
@@ -82,7 +82,7 @@ Alternatively, use the following example to AOT compile all assemblies, while st
 Another common scenario where the interpreter is sometimes required is a .NET MAUI Mac Catalyst app running on the ARM64 architecture, which can throw an exception on launch. This launch exception can often be fixed by enabling the interpreter:
 
 ```xml
-<PropertyGroup Condition="$(TargetFramework.Contains('-maccatalyst-arm64')) and '$(Configuration)' == 'Release'">
+<PropertyGroup Condition="'$(RuntimeIdentifier)' == 'maccatalyst-arm64' and '$(Configuration)' == 'Release'">
     <MtouchInterpreter>-all,MyAssembly.dll</MtouchInterpreter>
 </PropertyGroup>
 ```
--- a/docs/macios/interpreter.md
+++ b/docs/macios/interpreter.md
@@ -36,7 +36,7 @@ The Mono interpreter can be enabled in release builds by setting the `UseInterpr
 </PropertyGroup>
 ```
 
-The interpreter can be enabled for Mac Catalyst release builds on ARM64 by changing `-ios` in the above example to `-maccatalyst-arm64`.
+The interpreter can be enabled for Mac Catalyst release builds on ARM64 by changing `-ios` in the above example to `-maccatalyst`.
 
 > [!WARNING]
 > Don't enable the interpreter for release builds on Android because it disables JIT compilation.
--- a/docs/macios/interpreter.md
+++ b/docs/macios/interpreter.md
@@ -8,23 +8,23 @@ ms.date: 05/24/2023
 
 When you compile a .NET Multi-Platform App UI (.NET MAUI) app for iOS or Mac Catalyst, the compiler turns your app code into Microsoft Intermediate Language (MSIL). When you run the iOS app in the simulator, or the Mac Catalyst app, the .NET Common Language Runtime (CLR) compiles the MSIL using a Just in Time (JIT) compiler. At runtime the MSIL is compiled into native code, which can run on the correct architecture for your app.
 
-However, there is a security restriction on iOS, set by Apple, which disallows the execution of dynamically generated code on a device. Similarly, the execution of dynamically generated code is disallowed on ARM64-based Mac Catalyst apps. To meet this restriction, iOS and Mac Catalyst apps use an Ahead of Time (AOT) compiler to compile the managed code. This produces a native iOS binary that can be deployed to Apple devices, or a native Mac Catalyst binary.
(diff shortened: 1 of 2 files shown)


docs/macios/interpreter.md
```diff
@@ -0,0 +1,82 @@
...
+description: Learn how to enable the Mono interpreter, which lets you use dynamic code generation in your .NET MAUI iOS and ARM64-based Mac Catalyst release builds.
+ms.date: 05/24/2023
+---
+
+# Mono interpreter on iOS and Mac Catalyst
+
+When you compile a .NET Multi-Platform App UI (.NET MAUI) app for iOS or Mac Catalyst, the compiler turns your app code into Microsoft Intermediate Language (MSIL). When you run the iOS app in the simulator, or the Mac Catalyst app, the .NET Common Language Runtime (CLR) compiles the MSIL using a Just in Time (JIT) compiler. At runtime the MSIL is compiled into native code, which can run on the correct architecture for your app.
+
+However, there is a security restriction on iOS, set by Apple, which disallows the execution of dynamically generated code on a device. Similarly, the execution of dynamically generated code is disallowed on Mac Catalyst apps running on the ARM64 architecture. To meet this restriction, iOS and Mac Catalyst apps use an Ahead of Time (AOT) compiler to compile the managed code. This produces a native iOS binary that can be deployed to Apple devices, or a native Mac Catalyst binary.
```
rolfbjarne: Actually we have the same restriction when building for arm64 for the iOS simulator. Currently we always build for x64 when targeting the iOS simulator, so this rarely shows up, but we'll change the default for .NET 8 (and then we'll default to building for arm64 iOS simulator when on an arm64 Mac), so it'll become more common.

The solution is the same as for Mac Catalyst / arm64: use the interpreter.

docs/macios/interpreter.md
```diff
@@ -0,0 +1,82 @@
...
+
+The Mono interpreter can be enabled in release builds by setting the `UseInterpreter` MSBuild property to `true` in your .NET MAUI app's project file:
+
+```xml
+<PropertyGroup Condition="$(TargetFramework.Contains('-ios')) and '$(Configuration)' == 'Release'">
+    <UseInterpreter>true</UseInterpreter>
+</PropertyGroup>
+```
+
+The interpreter can be enabled for Mac Catalyst release builds on ARM64 by changing `-ios` in the above example to `-maccatalyst-arm64`.
```
rolfbjarne:
```suggestion
The interpreter can be enabled for Mac Catalyst release builds on ARM64 by changing `-ios` in the above example to `-maccatalyst`.
```

If you want to condition on arm64, you need to do something like this:

```xml
<PropertyGroup Condition="$(RuntimeIdentifier)' == 'maccatalyst-arm64' and '$(Configuration)' == 'Release'">
    <UseInterpreter>true</UseInterpreter>
</PropertyGroup>
```

That said, setting `UseInterpreter` on x64 is harmless, so it doesn't really matter.

PR #1561, opened 2023-06-24
Update Storage.cs

ZeProgFactory:
Changed platform from DevicePlatform.macOS to MacCatalyst.


davidbritch (approved):
Thanks @ZeProgFactory

PR #1584, opened 2023-07-05
Create version-sweep.yml GitHub Action workflow.

IEvangelist:
I'm proposing that the MAUI docs rely on this GitHub Action workflow as so many other docs repos do. It will run on the first of every month, and conditionally create an issue to help ensure that the sample source code projects remain up-to-date.

Contributes to https://github.com/dotnet/docs/issues/36108

### DIFF
--- a/.github/workflows/version-sweep.yml
+++ b/.github/workflows/version-sweep.yml
@@ -1,0 +1,47 @@
+# This is a basic workflow to help you get started with Actions
+
+name: "target supported version"
+
+# Controls when the action will run.
+on:
+  # Triggers the workflow on push or pull request events but only for the default branch
+  schedule:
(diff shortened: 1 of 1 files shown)


davidbritch:
@IEvangelist Thanks for this. Just double checking - it will check for latest (stable) version, right? There's no concept of LTS for .NET MAUI.

IEvangelist:
> @IEvangelist Thanks for this. Just double checking - it will check for latest (stable) version, right? There's no concept of LTS for .NET MAUI.

Yeah, so it will create issues for tracking the proposed upgrades of any out-of-support versions. It examines the `TargetFramework(s)` tag of the project file.

davidbritch (approved):

PR #1596, opened 2023-07-15
Update navigation.md : A highlighted line is missing

jgil83000:
Line 6 also has a change with changing the URL to "about.MoreInfoUrl"


davidbritch (approved):
Thanks @jgil83000

PR #1673, opened 2023-08-31
Add missing driverkit XML to entitlements-reference.md

Eilon:
Based on this doc: https://developer.apple.com/documentation/bundleresources/entitlements/com_apple_developer_driverkit_communicates-with-drivers?language=objc


davidbritch (approved):
Thanks @Eilon

PR #1991, opened 2023-12-28
Update AllNotesPage.xaml use theme colour for plus icon

danellisuk:
The plus symbol for adding a note was white which was not visible when using a light theme on Android.  This change uses the Primary theme colour which is visible on both light and dark mode.


davidbritch (approved):
Thanks @danellisuk

PR #2065, opened 2024-02-08
Fix inverted 'IgnoreSafeArea' value

Yves-Be:
The IgnoreSafeArea property needs to be set to 'true' to have any effect. The default is 'false', in which case the safe area is respected.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/ios/platform-specifics/page-safe-area-layout.md](https://github.com/dotnet/docs-maui/blob/f9d8abee9122803966b8050b9b3d5b3150d34252/docs/ios/platform-specifics/page-safe-area-layout.md) | [Disable the safe area layout guide on iOS](https://review.learn.microsoft.com/en-us/dotnet/maui/ios/platform-specifics/page-safe-area-layout?branch=pr-en-us-2065) |


davidbritch (approved):
Thanks @Yves-Be

PR #2069, opened 2024-02-09
Update min supported MacCatalyst versions

Redth:
The .NET MAUI templates currently specify macOS 13.1 as the minimum supported version, not 10.15.  It's very difficult to test older OS versions such as 10.15 which are no longer supported by Apple.  This should clarify the min supported version expectations a bit better.

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/supported-platforms.md](https://github.com/dotnet/docs-maui/blob/58906318e5651674f9d791b14839e1a0eb5d7a4b/docs/supported-platforms.md) | [Supported platforms for .NET MAUI apps](https://review.learn.microsoft.com/en-us/dotnet/maui/supported-platforms?branch=pr-en-us-2069) |


Redth:
I put 13.1, but that's really the MacCatalyst version I believe which isn't the _macOS_ version.  Investigating this a bit more, but given the [updated docs](https://github.com/dotnet/core/pull/9048) to be `12.0`+ I think this may align with MacCatalyst `15.0` instead.

docs/supported-platforms.md
```diff
@@ -10,14 +10,14 @@ ms.date: 04/13/2022
 
 - Android 5.0 (API 21) or higher.
 - iOS 11 or higher, using the latest release of Xcode.
```
akoeplinger: this is also iOS 12 now with .NET 9

Redth:
I think this is good now.

macOS 11 maps to MacCatalyst 14.0 [it seems](https://stackoverflow.com/questions/63581114/mac-catalyst-version)

Redth:
.NET 9 we will want to bump for MAUI apps to:
- iOS 13.0 
- macOS 12 (MacCatalyst 15.0)

davidbritch (approved):

PR #2147, opened 2024-03-14
add write permissions for pull requests

BillWagner:
GitHub permissions are ANDed, not ORed. So, even though the sequester app was given permission to read and write pull requests, the configuration in YAML prevented it.

This ensures that if a PR is tagged as a work item, it will be imported correctly.


davidbritch (approved):
LGTM

PR #2436, opened 2024-08-09
ActivateWindow method.

### LINKED ISSUE #2418: .NET MAUI 9: Bring Window to front
.NET MAUI 9 adds an `Application.ActivateWindow` method for Windows & Mac Catalyst.

More info: https://github.com/dotnet/maui/pull/23873

---
[Associated WorkItem - 295222](https://dev.azure.com/msft-skilling/Content/_workitems/edit/295222)

davidbritch:
Fixes #2418 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/fundamentals/windows.md](https://github.com/dotnet/docs-maui/blob/6bb9d19ef359104aeffd76affd47aa99aed4bbd1/docs/fundamentals/windows.md) | [docs/fundamentals/windows](https://review.learn.microsoft.com/en-us/dotnet/maui/fundamentals/windows?branch=pr-en-us-2436) |


docs/fundamentals/windows.md
```diff
@@ -111,19 +111,29 @@ Multiple windows can be simultaneously opened on Android, iOS on iPad (iPadOS),
 
 ```csharp
 Window secondWindow = new Window(new MyPage());
-Application.Current.OpenWindow(secondWindow);
+Application.Current!.OpenWindow(secondWindow);
```
jamesmontemagno: would it be better to do `Application.Current?.OpenWindow(secondWindow);`

I think we should avoid ! bad practice IMHO

davidbritch: Done!

MartyIX: > I think we should avoid ! bad practice IMHO

I disagree with this. `!` allows you to fail soon and discover issue sooner (fail-fast principle). `?` hides the fact that nothing happened (the app does not crash but it still contains a bug...).

jamesmontemagno: i mean probably we would want to check for null ahead of time. 

Using ? is only really for unit testing the code. If you are running the app... the app exists and it will not be null, so there is no failing or hiding anything.

MartyIX: > i mean probably we would want to check for null ahead of time.
 
Yes

> Using ? is only really for unit testing the code. 

If you happen to check MAUI's source code, you will find many `?`s in the actual product code. Mostly, it's similar to use of `var` which is different from project to project (no var, only var, both acceptable, etc.)

> If you are running the app... the app exists and it will not be null, so there is no failing or hiding anything.

In MauiProgram.cs, it will be null. In other files, it should be OK. Hence, `Application.Current!` should be correct in 99% of cases and it crashes when it should crash. `?` would "work" even in MauiProgram.cs and it would hide a bug.

PR #2537, opened 2024-10-03
Visual states examples need VisualStateGroupList for them to work properly

stephenquan:
When trying the visual states documented examples for Button and Entry they threw an exception.
Details of the crash is described in this issue https://github.com/dotnet/maui/issues/25069

A simple solution would be to update the visual states' documented examples for Button and Entry to include VisualStateGroupList.

---
[Associated WorkItem - 320968](https://dev.azure.com/msft-skilling/Content/_workitems/edit/320968)


davidbritch (approved):
@stephenquan I thought these had all been caught previously, so thanks for catching them - much appreciated.

PR #2607, opened 2024-11-07
Unpackaged app deployment

### LINKED ISSUE #2415: .NET MAUI 9: Update doc to describe how .NET 9 uses unpackaged F5 to Windows
### Type of issue

Other (describe below)

### Description

For .NET 9, we're changing the default F5 deployment for Windows to Unpackaged so that we can address new users who are unable to turn on Developer Mode on their Windows machines. This will be a change for new projects that will add the WindowsPackageType attribute in the .csproj and set the launchSetting.json commandName to "Project". We need to document how to switch it back to Packaged if users need to access Windows Deployment APIs that only work with Packaged apps. 

PR is here: https://github.com/dotnet/maui/pull/23787 

### Page URL

https://learn.microsoft.com/en-us/dotnet/maui/windows/setup?view=net-maui-9.0

### Content source URL

https://github.com/dotnet/docs-maui/blob/main/docs/windows/setup.md

### Document Version Independent Id

e139fd32-4027-b014-115f-18f53ab6a46c

### Article author

@davidbritch

### Metadata

* ID: e139fd32-4027-b014-115f-18f53ab6a46c 
* Service: **dotnet-mobile**
* Sub-service: **dotnet-maui**

---
[Associated WorkItem - 335728](https://dev.azure.com/msft-skilling/Content/_workitems/edit/335728)

davidbritch:
Fixes #2415 

---

#### Internal previews

| 📄 File | 🔗 Preview link |
|:--|:--|
| [docs/whats-new/dotnet-9.md](https://github.com/dotnet/docs-maui/blob/02574506d302550e379dbaa3ded5f79dcd39b132/docs/whats-new/dotnet-9.md) | [docs/whats-new/dotnet-9](https://review.learn.microsoft.com/en-us/dotnet/maui/whats-new/dotnet-9?branch=pr-en-us-2607) |
| [docs/windows/setup.md](https://github.com/dotnet/docs-maui/blob/02574506d302550e379dbaa3ded5f79dcd39b132/docs/windows/setup.md) | [docs/windows/setup](https://review.learn.microsoft.com/en-us/dotnet/maui/windows/setup?branch=pr-en-us-2607) |


davidbritch:
@BethMassi Two questions:

1. Is the ability to switch between packaged and unpackaged .NET 9 only, or also .NET 8? At the moment I limited that content in the doc to .NET 9.
1. Maybe the new content should go at the top of the doc? I've tried it both ways around and I keep changing my mind on what works best. We can, if required, have different orderings in the .NET 8 and .NET 9 versions of the doc.

BethMassi:
Hi @davidbritch 
The ability to switch in the VS IDE exists for .NET 8 and 9 projects (VS version 17.12). However, only .NET 9 projects default to Unpackaged. .NET 8 projects are unchanged (default to Packaged). I also think it's helpful to show the code changes that the IDE actually makes in the .csproj and launchSetting.json. If someone has defined multiple launch settings, then the IDE switch will not touch those entries.

BethMassi:
> 2. Maybe the new content should go at the top of the doc? I've tried it both ways around and I keep changing my mind on what works best. We can, if required, have different orderings in the .NET 8 and .NET 9 versions of the doc.

I think it would be better in the .NET 9 version to jump right to the Windows debug target (since you get Unpackaged by default) and then show folks how they can switch to packaged if needed.

davidbritch:
@BethMassi Should be good now.

docs/windows/includes/target-windows.md
```diff
@@ -0,0 +1,10 @@
+---
+ms.topic: include
+ms.date: 11/07/2024
+---
+
+## Target Windows
+
+In Visual Studio, set the **Debug Target** to **Framework (...)** > **net8.0-windows**. There is a version number in the item entry, which may or may not match the following screenshot:
```
BethMassi, line 10: This should say net9.0-windows for .NET 9 version of the doc.

BethMassi (approved):
Looks great, just one suggestion.

docs/windows/includes/target-windows.md
davidbritch: Yep, I've got a separate task to update net8-*, but I might as well fix it here now.

PR #2643, opened 2024-11-25
linker-control: IsAotCompatible implies IsTrimmable.

breyed:
Copied the text from prepare-libraries-for-trimming.md that provides guidance for when IsAotCompatible is true.


davidbritch (approved):
Thanks @breyed

PR #2667, opened 2024-12-06
Update create.md

mr5z:
Add auto type cast within pattern matching occurences


mr5z:
@dotnet-policy-service agree

davidbritch:
Hi @mr5z 

To keep the samples and docs sync'd, could you also update the sample code at https://github.com/dotnet/maui-samples/tree/main/9.0/UserInterface/Handlers/CreateHandlerDemo?

PR #2672, opened 2024-12-10
Update remote-simulator.md with workaround steps for boot loops

drasticactions:
Xcode can generate simulators with null data that causes the iOS Remote Simulator to trip up with an NRE. This has been addressed internally and can also be worked around by removing and recreating the simulators.


davidbritch (approved):
Thanks @drasticactions

PR #2720, opened 2025-01-10
Update single-project.md | using correct extension method ToPlatform() instead of ToNative()

bhavanesh2001:
In the example of Platform-specific code section `ToNative()` extension method is being used on a Color. 

But In reality, There is no such extension method of Color exists.

It need's to replaced with `ToPlatform()` extension method which exists.


bhavanesh2001:
#sign-off

davidbritch (approved):

PR #2724, opened 2025-01-12
Update webview.md

djukic-marko:
While trying to make it work on my device, i was constantly appedning .Platforms.Android path, since it was logical. However, it should not be appended, and could save someones time.


davidbritch (approved):
Thanks @djukic-marko

djukic-marko:
Thanks for listening to feedback. I'll do a better job with grammar next time!

PR #2736, opened 2025-01-16
Adding a Tip to specify ScrollTo may not work under specific condition's

bhavanesh2001:
As discussed [here ](https://github.com/dotnet/maui/issues/4115). The ScrollTo method may not work depending up on the layout it's wrapped up in.

This behavior should be highlighted in the documentation to prevent confusion.

### DIFF
--- a/docs/user-interface/controls/collectionview/includes/scrolling-tip.md
+++ b/docs/user-interface/controls/collectionview/includes/scrolling-tip.md
@@ -1,7 +1,7 @@
 ---
 ms.topic: include
-ms.date: 06/09/2023
+ms.date: 01/17/2025
 ---
 
 > [!TIP]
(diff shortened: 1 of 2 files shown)


bhavanesh2001:
Note that there is already a similar tip defined [here](https://github.com/dotnet/docs-maui/blob/050cd803c7ac9e8347142de3abd2375331d721cd/docs/user-interface/controls/collectionview/scrolling.md?plain=1#L23) but it didn't mention anything about **ScrollTo** method. It refers to physical scrolling by user.

bhavanesh2001:
@dotnet-policy-service agree

davidbritch (approved):
Thanks @bhavanesh2001 

Lots of lovely xrefs :D

PR #2792, opened 2025-02-17
Increase the minimum apple versions

bijington:
This increases the minimum versions provided by the .NET MAUI template at: https://github.com/dotnet/maui/blob/main/src/Templates/src/templates/maui-mobile/MauiApp.1.csproj

We discovered this in an issue for the toolkit and wanted to help bring this up to date.
https://github.com/CommunityToolkit/Maui/issues/2528

### DIFF
--- a/docs/supported-platforms.md
+++ b/docs/supported-platforms.md
@@ -26,8 +26,8 @@ ms.date: 10/04/2024
 ::: moniker range=">=net-maui-9.0"
 
 - Android 5.0 (API 21) or higher is required.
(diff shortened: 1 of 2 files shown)


davidbritch:
Hi @bijington 

So I looked at this at the time .NET 9 was released, and my understanding IIRC is that 15.0 is an attempt to force people to adopt later versions, but that the actual minimum supported version hasn't changed.

bijington:
Oh that is interesting! Thanks

PR #2808, opened 2025-02-27
- fix android setup for enableing Developer options

BrayanKhosravian:


davidbritch (approved):
Thanks @BrayanKhosravian

BrayanKhosravian:
And thanks for the review, @davidbritch.
I believe it’s more efficient to propose changes directly via pull requests rather than opening an issue first.

PR #2831, opened 2025-03-12
Add new Blazor Hybrid article to the ToC

guardrex:
Hello (David?)!

@BethMassi and I would like to add the new .NET MAUI with Blazor Web App and Identity sample article in the Blazor docs to the ToC.


davidbritch (approved):
Thanks @guardrex

PR #2881, opened 2025-04-21
Pair to Mac docs update

emaf:
- Add SSH Keys generation section to the Pair to Mac docs explaining how to get keys compatible with both Xamarin and MAUI
- Rename the remote setup section


davidbritch (approved):

rachelkang:
Hi @davidbritch - the final version that got merged here is a bit misleading, particularly:

> SSH keys generated when running against macOS 15.4.1+ are not compatible with .NET for iOS. If you're working on both .NET MAUI and .NET for iOS projects, and use different macOS versions depending on the project type, ensure you perform the initial connection using a macOS version prior to 15.4.1 to ensure that you have an SSH key compatible with both type of projects.

They are actually compatible with .NET for iOS, but **not compatible with Xamarin.iOS**. This note is meant to address the situation of a migrating user who needs to Pair to Mac from both their old Xamarin project and new .NET MAUI project. As such, we'd like to make that revision, but also open to moving this info elsewhere if you believe the Xamarin mention makes it inappropriate to have here

PR #2887, opened 2025-04-24
Fix heading levels in C# sample

mscottford:
The C# sample did not match the XAML sample. This change should make the two samples equivalent.


davidbritch (approved):
Thanks @mscottford

Issue #2885, opened 2025-04-23
android.hardware.camera is missing XML manifest documentation


IeuanWalker:
### Type of issue

Missing information

### Description

You have the following documented for assembly attributes -
```csharp
// Add these properties if you would like to filter out devices that do not have cameras, or set to false to make them optional
[assembly: UsesFeature("android.hardware.camera", Required = true)]
[assembly: UsesFeature("android.hardware.camera.autofocus", Required = true)]
```

The equivalent for the manifest XML is missing -
```xml
<!-- Add these properties if you would like to filter out devices that do not have cameras, or set to false to make them optional -->
<uses-feature android:name="android.hardware.camera" android:required="true"/>
<uses-feature android:name="android.hardware.camera.autofocus" android:required="true"/>
```

____

It would probably be better as well if the assembly and the manifest were ordered/ commented the same as well _(assembly is clearer)_

### Page URL

https://learn.microsoft.com/en-us/dotnet/maui/platform-integration/device-media/picker?view=net-maui-9.0&tabs=android

### Content source URL

https://github.com/dotnet/docs-maui/blob/main/docs/platform-integration/device-media/picker.md

### Document Version Independent Id

1e5277f4-f2e4-dffe-0c08-06531d4145a1

### Platform Id

27207929-eaa2-6859-b3f3-4f8a3728d85f

### Article author

@davidbritch

### Metadata

* ID: 1e5277f4-f2e4-dffe-0c08-06531d4145a1
* PlatformId: 27207929-eaa2-6859-b3f3-4f8a3728d85f 
* Service: **dotnet-mobile**
* Sub-service: **dotnet-maui**

[Related Issues](https://github.com/dotnet/docs-maui/issues?q=is%3Aissue+is%3Aopen+1e5277f4-f2e4-dffe-0c08-06531d4145a1)

jfversluis:
I think this is fixed now with #3152

PR #2889, opened 2025-04-28
Add missing XML documentation for android.hardware.camera

IeuanWalker:
#2885

### DIFF
--- a/docs/platform-integration/device-media/picker.md
+++ b/docs/platform-integration/device-media/picker.md
@@ -46,9 +46,12 @@ These permissions can be added in the following ways:
   ```xml
   <!-- Needed for Picking photo/video -->
   <uses-permission android:name="android.permission.WRITE_EXTERNAL_STORAGE" android:maxSdkVersion="32" />
-  <uses-permission android:name="android.permission.READ_MEDIA_AUDIO" /> <!-- Required only if your app needs to access audio files that other apps created -->
-  <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" /> <!-- Required only if your app needs to access images or photos that other apps created -->
-  <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" /> <!-- Required only if your app needs to access videos that other apps created -->
+  <!-- Required only if your app needs to access audio files that other apps created -->
+  <uses-permission android:name="android.permission.READ_MEDIA_AUDIO" />
+  <!-- Required only if your app needs to access images or photos that other apps created -->
+  <uses-permission android:name="android.permission.READ_MEDIA_IMAGES" />
+  <!-- Required only if your app needs to access videos that other apps created -->
+  <uses-permission android:name="android.permission.READ_MEDIA_VIDEO" />
 
   <!-- Needed for Taking photo/video -->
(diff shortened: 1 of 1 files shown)


adegeo (dismissed):
I like the idea of ordering them in the same way, but the XML comment isn't easily visible when reading the article. Restoring the XML comment above the setting helps readability. Maybe even adding a blank line in there between items.

Perhaps match the ordering in the block at the top of the article?

I also noticed that the linked issue doesn't seem to match what you're changing? In that you talk about the camera?

IeuanWalker:
@adegeo no prob, will take a look tomorrow.

The linked issue is about these 2 lines being missing, which have been added - 
image

IeuanWalker:
> I like the idea of ordering them in the same way, but the XML comment isn't easily visible when reading the article. Restoring the XML comment above the setting helps readability. Maybe even adding a blank line in there between items.
> 
> The block at the top of the article also be set to match.
> 
> I also noticed that the linked issue doesn't seem to match what you're changing? In that you talk about the camera?

@adegeo  I've moved the comments back above each option.

___________________

I haven't added _"a blank line in there between items"_, because currently the blank lines indicate the permission for the different features - 
image


Not sure what you meant by _"The block at the top of the article also be set to match"_

adegeo:
Wow, I don't know how I ended up wording that sentence in that way. I think I spliced two sentences together while editing and didn't realize how awful it sounded. HA!

I edited my comment to fix the grammar. I was referencing the ordering in the article above where the change was made, that ordering could match the ordering in the xml:

image

EDIT:
After looking a bit more at this, and the published versions (original vs updated) I don't think I like this change because it makes it unclear what's required specifically for the **picker**. Sure, in your actual app you may want to structure the XML better than this sample, but this sample is really tuned to show that:

- READ and WRITE are required for Android 12
- The individual items are required for Android 13.

Having them all listed as the original version did, makes that clear.

IeuanWalker:
But in that case shouldnt both `assembly` and `manifest` versions be laid out in the same way.

image 

Personally i think the `assembly`  is more clearly laid out compared to the `manifest` version, especially if your only using part of the feature and don't need all the permissions etc. 

So this PR just updates the `manifest` to match the `assembly` version, and adds the missing XML, i.e. - 
```xml
  <!-- Add these properties if you would like to filter out devices that do not have cameras, or set to false to make them optional -->
  <uses-feature android:name="android.hardware.camera" android:required="true"/>
  <uses-feature android:name="android.hardware.camera.autofocus" android:required="true"/>
```

adegeo:
With the recent shifts we've had around here, someone from @davidortinau team will have to take this over. Sorry about that. I'll dismiss my review and let them evaluate this. Again, sorry for the slow traction on this.

jfversluis:
Goodness @IeuanWalker I'm so sorry, you had this contribution since forever and I really wanted to take this one. Somehow I mistook the recently opened #3152 and merged that one instead. Honest mistake, my sincere apologies. Also for taking so long to get around to this. 

We will be monitoring this more closely and I will personally promise you I will look at your contributions faster should you still be willing to make any.

Happy new year!

IeuanWalker:
@jfversluis no prob, happy new year :)

PR #2957, opened 2025-07-30
Document NuGet reference change when using CPM

### LINKED ISSUE #30881: Resizetizer no longer implicitly referenced for MAUI class libraries
### Description

This may be intended behavior, but I still wanted to put it here so the next guy can search for the issue.

When using assets (`MauiAsset`) or images (`MauiImage`) in a class library, .NET 8 would automatically generate the necessary build options & outputs if the `UseMaui` flag was set to `true`. In .NET 9, this is no longer the case: The library project must now implicitly or explicitly reference `Microsoft.Maui.Resizetizer`. Otherwise, none of the project's assets are included in the output.

### Expected behavior
Any MAUI project's assets should be usable in the application. Required dependencies for preprocessing should be pulled in automatically.

This previously worked in 8.0.100. I have not tried older 9.* versions.

### Steps to Reproduce

- Create two .NET 9 projects: The main application project (I used `net9.0-android` TFM, but I believe this occurs for other platforms as well) and a MAUI class library project (`net9.0` TFM).
- Enable CPM (`<ManagePackageVersionsCentrally>true</ManagePackageVersionsCentrally>`).
- Add an SVG file to the library. Reference the file in the project using `<MauiImage Include="my-image.svg" />`.
- Try to display the converted image in the application, e.g. with `<Image Source="my-image.png" />`.
- Observe how the image is not displayed. Examining the binary logs reveals that the Resizetizer and related build actions were not imported and did not run.

### Link to public reproduction project repository

_No response_

### Version with bug

9.0.90 SR9

### Is this a regression from previous behavior?

Yes, this used to work in .NET MAUI

### Last version that worked well

.NET 8  (Please specify exact version)

### Affected platforms

Android

### Affected platform versions

Tested with Android 14, but seems to affect all platforms.

### Did you find any workaround?

Manually add a package reference for Resizetizer: `<PackageReference Include="Microsoft.Maui.Resizetizer" />`.

### Relevant log output

```shell

```

DJ4ddi:
Fixes https://github.com/dotnet/maui/issues/30881

### DIFF
--- a/docs/whats-new/dotnet-8.md
+++ b/docs/whats-new/dotnet-8.md
@@ -176,7 +176,7 @@ Explicit package references should also be added to your *.csproj* file for the 
 </ItemGroup>
 ```
 
-When using [Central Package Management (CPM)](/nuget/consume-packages/central-package-management), MAUI packages are no longer referenced automatically and must be explicitly added. For projects that only contain assets (`MauiImage`, `MauiAsset`, etc.), it is sufficient to add the `Microsoft.Maui.Resizetizer` package (instead of `Microsoft.Maui.Controls`).
+When using [Central Package Management (CPM)](/nuget/consume-packages/central-package-management), .NET MAUI packages are no longer referenced automatically and must be explicitly added. For projects that only contain assets (`MauiImage`, `MauiAsset`, etc.), it is sufficient to add the `Microsoft.Maui.Resizetizer` package (instead of `Microsoft.Maui.Controls`).
 
 The `$(MauiVersion)` variable is referenced from the version of .NET MAUI you've installed. You can override this by adding the `$(MauiVersion)` build property to your *.csproj* file:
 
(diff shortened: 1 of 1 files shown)


docs/whats-new/dotnet-8.md
```diff
@@ -175,6 +176,8 @@ Explicit package references should also be added to your *.csproj* file for the
 </ItemGroup>
 ```
 
+When using [Central Package Management (CPM)](/nuget/consume-packages/central-package-management), MAUI packages are no longer referenced automatically and must be explicitly added. For projects that only contain assets (`MauiImage`, `MauiAsset`, etc.), it is sufficient to add the `Microsoft.Maui.Resizetizer` package (instead of `Microsoft.Maui.Controls`).
```
jfversluis:
```suggestion
When using [Central Package Management (CPM)](/nuget/consume-packages/central-package-management), .NET MAUI packages are no longer referenced automatically and must be explicitly added. For projects that only contain assets (`MauiImage`, `MauiAsset`, etc.), it is sufficient to add the `Microsoft.Maui.Resizetizer` package (instead of `Microsoft.Maui.Controls`).
```

jfversluis:
Appreciate the help here, thank you!

jfversluis (approved):

PR #2972, opened 2025-08-05
Add workaround for Windows line endings on the Editor control

### LINKED ISSUE #2849: Editor control uses \r line endings on Windows (not expected \r\n)
## Documentation

Please add to the documentation of the Editor control that it uses `\r` line endings on Windows. One would expect it to be `\r\n`, which is the normal line ending for Windows. Also please provide an example how to convert line endings from `\r` to `\r\n`. Something like this:

```
#if WINDOWS
    string textWithWindowsLineEndings = Editor1.Text.Replace("\r", "\r\n");
#endif
```

## Feature

One could also think about adding a feature for doing this automatically to .NET MAUI.

## Related Info
https://github.com/microsoft/microsoft-ui-xaml/issues/1826

JohanSmarius:
Fix #2849

### DIFF
--- a/docs/user-interface/controls/editor.md
+++ b/docs/user-interface/controls/editor.md
@@ -255,12 +255,12 @@ When the user adds multiple lines in the control, the `Text` property will conta
 
 ```csharp
 #if WINDOWS
-		if (EditorText.Text.Contains("\r"))
-		{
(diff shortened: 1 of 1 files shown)


jfversluis (approved):
I keep saying it: thank you!

PR #3009, opened 2025-08-28
Update navigationpage.md

educfn:
Correcting minor errors


jfversluis (approved):
Oh good catch! Thank you for that!

PR #3078, opened 2025-11-10
Revise command interface usage for IsEnabled property

kkilpatrickjjk:
Updated documentation to state that the `IsEnabled` property can be used with command-based buttons from .NET MAUI 8 and above. Explained the implications thereof as well.

The `IsEnabled` property's non-functional nature with command-based buttons was corrected all the way back in [.NET MAUI preview 8.0.0-preview.1.7762](https://github.com/dotnet/maui/releases/tag/8.0.0-preview.1.7762) with MAUI pull request [12488](https://github.com/dotnet/maui/pull/12488), but this documentation was not updated at that time.


kkilpatrickjjk:
I'm a little confused about the error message here. I did not delete a branch or do a force push. And the linked instructions to resolve the issues just get me a 404.

jfversluis:
@kkilpatrickjjk must have been some hiccup in the system, seems all good now, thanks for this! The one thing I do think about this one is: .NET MAUI 8 is out of support... That means .NET MAUI 7 has been out of support even longer, so how relevant is this still? But I guess since you're running into it it might still be... And it doesn't really hurt to have it in here for the time being so people who stumble on it will get value from it.

jfversluis (approved):

kkilpatrickjjk:
@jfversluis I had wondered the same thing about whether it was worth mentioning given how long they'd been out of support. My thought was that as there are probably still people like me who were taking that message about "command-based buttons don't work" as true even when it wasn't, and as there are Stack Overflow answers out there citing the old version of the truth (and as AIs answering the question for people might thus come to that now-incorrect conclusion), that that makes it important to have an official message both acknowledging that well-documented previous state and setting the record straight about current capabilities. Something to serve as a bridge from those popular old answers to the newer state of things

Thank you for your review, approval, and merge!

PR #3137, opened 2025-12-13
Update include syntax for picker documentation

matt-goldman:
There's a small error in the Picker page: https://learn.microsoft.com/en-us/dotnet/maui/user-interface/controls/picker?view=net-maui-10.0

image

Looking at other examples I think this is the right fix

### DIFF
--- a/docs/user-interface/controls/picker.md
+++ b/docs/user-interface/controls/picker.md
@@ -262,6 +262,6 @@ This method obtains the `SelectedIndex` property value, and uses the value to re
 
 ::: moniker range=">=net-maui-10.0"
 
-::: include ../../includes/pickers-open-close-dotnet10.md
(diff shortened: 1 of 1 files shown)


matt-goldman:
Looks like the same issue affects `DatePicker` and `TimePicker`. The include file doesn't actually exist. Was this a placeholder?

matt-goldman:
@davidortinau looks like you made this change here: 05d273b194b39477a8b6dc04643efb6c0af39f4b

Apologies, I'm missing something here and went for a knee-jerk fix without looking through this thoroughly (in my defence it's 6am on a Sunday here!). But FYI these docs are not rendering properly. I'll close this and open an issue instead.

jfversluis:
Looks like you were going up a level too much 😄 #3141 thanks for the brave attempt at such an unholy hour!

PR #3270, opened 2026-03-25
Pan gesture: fix XAML example to use Grid and dotnet_bot.png

### LINKED ISSUE #3251: Pan gesture: document that TotalX/TotalY semantics differ between iOS (cumulative) and Android (incremental)
## Summary

The pan gesture documentation describes `TotalX` and `TotalY` properties without noting that their meaning differs between platforms. On iOS, these values are cumulative (relative to where the gesture started). On Android, they are incremental (relative to the previous event). This silent behavioral difference causes pan gesture implementations that work correctly on one platform to behave incorrectly on the other.

## Why it matters

A developer testing on iOS will write code like:

```csharp
pan.PanUpdated += (s, e) =>
{
    if (e.StatusType == GestureStatus.Running)
    {
        image.TranslationX = e.TotalX;  // works on iOS — cumulative from start
        image.TranslationY = e.TotalY;
    }
};
```

This code produces smooth, correct movement on iOS. On Android, `TotalX`/`TotalY` represent the delta since the *last event*, so directly assigning them causes the view to jump to a tiny offset each frame instead of tracking the full pan distance — a clearly broken UX that only manifests on Android.

The inverse is also true: code written for Android (accumulating deltas manually) will double-count on iOS.

## What should be documented

Add a platform behavior note to `docs/fundamentals/gestures/pan.md` explaining the coordinate difference and providing a normalized pattern:

| Platform | `TotalX` / `TotalY` semantics |
|----------|-------------------------------|
| iOS | Cumulative offset from the **gesture start position** |
| Android | Incremental offset from the **previous event** |
| Windows | Cumulative offset from the **gesture start position** |

To write cross-platform pan code, track the previous translation and accumulate deltas on Android, or use the cumulative values directly on iOS/Windows:

```csharp
double _startTranslationX, _startTranslationY;

pan.PanUpdated += (s, e) =>
{
    switch (e.StatusType)
    {
        case GestureStatus.Started:
            _startTranslationX = image.TranslationX;
            _startTranslationY = image.TranslationY;
            break;
        case GestureStatus.Running:
            image.TranslationX = _startTranslationX + e.TotalX;
            image.TranslationY = _startTranslationY + e.TotalY;
            break;
        case GestureStatus.Completed:
            _startTranslationX = image.TranslationX;
            _startTranslationY = image.TranslationY;
            break;
    }
};
```

This pattern works correctly on all platforms because it captures the starting translation and applies `TotalX`/`TotalY` as an offset from that baseline.

## Suggested location

`docs/fundamentals/gestures/pan.md` — in the `PanGestureRecognizer` section, near the `TotalX`/`TotalY` property description

sheiksyedm:
- Replace AbsoluteLayout with Grid in the PanContainer XAML example
- Replace monkey.jpg (with fixed WidthRequest/HeightRequest) with dotnet_bot.png
- Update ms.date to 03/24/2026

Fixes #3251

### DIFF
--- a/docs/fundamentals/gestures/pan.md
+++ b/docs/fundamentals/gestures/pan.md
@@ -1,7 +1,7 @@
 ---
 title: "Recognize a pan gesture"
 description: "This article explains how to use a .NET MAUI pan gesture to horizontally and vertically pan an image, so that all of the image content can be viewed when it's being displayed in a viewport smaller than the image dimensions."
-ms.date: 06/19/2023
+ms.date: 03/24/2026
 ---
 
 # Recognize a pan gesture
@@ -98,11 +98,11 @@ The `PanContainer` class can be wrapped around a <xref:Microsoft.Maui.Controls.V
              xmlns:x="http://schemas.microsoft.com/winfx/2009/xaml"
              xmlns:local="clr-namespace:PanGesture"
              x:Class="PanGesture.MainPage">
-    <AbsoluteLayout>
+    <Grid>
         <local:PanContainer>
-            <Image Source="monkey.jpg" WidthRequest="1024" HeightRequest="768" />
+            <Image Source="dotnet_bot.png" />
         </local:PanContainer>
-    </AbsoluteLayout>
+    </Grid>
 </ContentPage>
 ```
 


jfversluis (changes requested):
I don't see how this addresses the linked issue?

sheiksyedm:
> #3251

@jfversluis  The linked doc issue related to "android is not returning the cumulative offset from the gesture start position" have been addressed in the latest MAUI version through PR [#21547](https://github.com/dotnet/maui/pull/21547). Android behavior is now consistent with the other platforms, so there is no need to address the linked issue.

While verifying the sample with the documented code, I noticed that it was not working correctly. As a result, I updated the sample with a working code in this PR

jfversluis (approved):

PR #3419, opened 2026-07-18
Native embedding: document iOS app extension scenario

kiddailey:
## Description

.NET MAUI's native embedding feature (`UseMauiEmbeddedApp<TApp>()`) assumes
a real `UIApplication`/`AppDelegate`/`SceneDelegate` exists, since it
resolves its required `IUIApplicationDelegate` from
`UIApplication.SharedApplication.Delegate`. This assumption doesn't hold
inside an iOS app extension (for example, an `ASCredentialProviderExtension`
used to implement AutoFill password/passkey support), which runs as its
own short-lived process and never calls `UIApplicationMain`. As a result,
`UseMauiEmbeddedApp<TApp>()` throws `InvalidOperationException: Platform
application instance is required and was not able to be detected` when
called from an app extension.

This PR documents the workaround: supplying a minimal stand-in
`IUIApplicationDelegate` before building the `MauiApp`, and presenting
embedded content via a manually-created `UIViewController` instead of a
`SceneDelegate`-owned `UINavigationController`, since app extensions have
no scene of their own.

This is added as a new subsection, "Embed in an iOS app extension", under
the existing "Window context" section on the native embedding page,
scoped to `>=net-maui-9.0` and the `devices-ios, devices-maccatalyst` zone
pivot, consistent with the rest of the page's structure.

## Verification

This approach was verified against a working .NET MAUI iOS AutoFill
credential provider extension prototype: the extension registers as an
AutoFill provider, presents a full MAUI page inside the extension process,
and successfully autofills a credential into a real password field in
Safari on the iOS Simulator.

## Related

None of the neither behavior described here (the `null` delegate, or the
workaround) appears to be documented elsewhere currently.

### DIFF
--- a/docs/platform-integration/native-embedding.md
+++ b/docs/platform-integration/native-embedding.md
@@ -1113,6 +1113,65 @@ public sealed partial class MainWindow : Microsoft.UI.Xaml.Window
 
 In this example, the <xref:Microsoft.Maui.Hosting.MauiApp> object is created as a shared, static instance. When this object is created, `MauiProgram.CreateMauiApp` is called which in turn calls the `UseMauiEmbedding` extension method on the <xref:Microsoft.Maui.Hosting.MauiAppBuilder> object. Therefore, your native app project should include a reference to the .NET MAUI class library project you created that contains your `MauiProgram` class and your .NET MAUI UI. A <xref:Microsoft.Maui.MauiContext> object is then created with the `CreateEmbeddedWindowContext` method, that's scoped to the window. The <xref:Microsoft.Maui.MauiContext> object will be used when converting .NET MAUI controls to native types.
 
+:::zone pivot="devices-ios, devices-maccatalyst"
+
+### Embed in an iOS app extension
+
+The window context approach described above assumes your native app has a real `UIApplication`, `AppDelegate`, and `SceneDelegate` &mdash; that is, a normal app launch sequence. This doesn't hold for an iOS app extension, such as an [`ASCredentialProviderExtension`](/dotnet/api/foundation) used to implement AutoFill password/passkey support. An app extension runs as its own short-lived process, hosted directly by the system (for example, by `PlugInKit`), and never calls `UIApplicationMain`. As a result, `UIApplication.SharedApplication.Delegate` is always `null` in an app extension process.
+
+Native embedding still works in this scenario, but two adjustments are required beyond the standard window context walkthrough:
+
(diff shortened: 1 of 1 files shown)


jfversluis (changes requested):
Thank you @kiddailey for your contribution and for taking the time to document this native embedding scenario.

I don't think this workaround is safe for us to publish as supported guidance yet. Apple explicitly states that an app extension [cannot access the `sharedApplication` object](https://developer.apple.com/library/archive/documentation/General/Conceptual/ExtensibilityPG/ExtensionOverview.html#//apple_ref/doc/uid/TP40014214-CH2-SW6), and also notes that the App Store rejects extensions that use APIs unavailable to app extensions in its guidance on [using embedded frameworks](https://developer.apple.com/library/archive/documentation/General/Conceptual/ExtensibilityPG/ExtensionScenarios.html#//apple_ref/doc/uid/TP40014214-CH21-SW3). The proposed code directly accesses and assigns `UIApplication.SharedApplication.Delegate`, so success in the iOS Simulator doesn't establish that this is extension-safe or acceptable for App Store distribution.

The current MAUI implementation does resolve the platform delegate through [`UIApplication.SharedApplication.Delegate`](https://github.com/dotnet/maui/blob/a4ede427ec2812ac85fb72eabc61e4bdaac3c071/src/Core/src/Embedding/EmbeddingExtensions.cs#L31-L44), but that points to a framework support gap rather than a documentation-only workaround. Please defer or remove this guidance until MAUI provides or explicitly confirms a supported, extension-safe initialization path that doesn't access `SharedApplication`. An upstream MAUI issue or API change would be the appropriate place to establish that support first.

One smaller correction: `ASCredentialProviderExtension` isn't an API type, and the current link points to the unrelated `Foundation` namespace. Please use descriptive text such as "credential provider extension," or link to Apple's [`ASCredentialProviderViewController`](https://developer.apple.com/documentation/authenticationservices/ascredentialproviderviewcontroller) documentation.

kiddailey:
Thanks for the thorough review — you're right on both points, and since the whole PR is this one subsection, I'll close it rather than try to salvage part of it.

On the API link: agreed, `ASCredentialProviderExtension` isn't a real type. Noted for if/when I bring this back in another form.

On the `SharedApplication.Delegate` workaround: you're correct that this isn't safe to publish as guidance. I had totally forgot about Apple's app-extension restrictions on touching `UIApplication.sharedApplication`. Apologies.

We're pinned to 9.0 for the foreseeable future, so we'll follow your suggestion and file a proper feature request against dotnet/maui for a public, extension-safe overload, rather than reintroducing this as documented guidance here.

Closing this PR for now. Thanks again for catching this.

PR #3436, opened 2026-07-29
Update docs-verifier action and add env variables

gewarren:
Updated the docs verifier action to a new version and added environment variables for GitHub PR number and token.


wadepickett (approved):
@gewarren:  Approved.

